PR #11, opened 2019-01-21
make ParaTree class destructor virtual

pkestene:
ParaTree is a base class derived in PabloUniform; its destructor needs to be virtual


andrea-iob:
Thanks for the fix. May I ask you to add the "PABLO: " prefix to the commit message (i.e., "PABLO: make ParaTree class destructor virtual")? After that change I can merge your branch into master and 1.5 branches.

andrea-iob:
The fix is now in the master branch. @pkestene You will not find your commit on the master branch: at the time you discovered the problem it was already fixed on our internal development branch (I think we made the same fix at the beginning of January). Sorry for that. Thanks again for your contribution.

PR #21, opened 2019-09-09
Fix in getY/Z in PabloUniform

LucaVerzeroliUnibg:
Copy and paste bug probably. m_origin[0] is equal for X, Y and Z.


edoardolombardi:
Thank you Luca for your fixing. We'll merge the pull request as soon as possible.

andrea-iob:
I pushed the fix to the master branch (I used a slightly different commit message). We are currently working on the bitpit 1.6 release, the fix will be included in that version.

PR #26, opened 2019-12-06
PABLO: fix markers communication

marcocisternino:
SendBuffer of marker communicator is now get by reference and not by copy.


andrea-iob:
Please remove from the branch the commits unrelated to the fix.

PR #29, opened 2019-12-11
patchkernel: introduce LineKernel and PointKernel classes

andrea-iob:
The class LineKernel allows to define patches that have a co-dimension equal to two with respect to space of a patch (for a three-dimensional patch the LineKernel may be used to define segments, for a two-dimensional patch the LineKernel may be used to define points).

The class PointKernel allows to define patches that have a co-dimension equal to three with respect to space of a patch (for a three-dimensional patch the LineKernel may be used to define points, whereas it doesn't make sens to define a PointKernel for a two-dimensional patch).


andrea-iob:
@roccoarpa @edoardolombardi Do you have an estimate on when we can do the merge?

roccoarpa:
@andrea-iob @edoardolombardi sorry for that, we will have meetings in the next few days to update priorities on mimmo. I will try to insert this pull check (coming with major mimmo rework) on them as urgent task, but I'm afraid is not the top priority in the short period. Just to answer your question, I will say 2/3 months from now as reasonable deadline, maybe earlier if some time slots get free in the meantime.  I cannot be more specific right now.

andrea-iob:
This pull request is here from a very long time. If it's not possible to merge the whole pull request, can we merge at least the commits that introduce the LineUnstructured and PointClound modules? It would be good to merge also the commit remove the space dimension parameter from the SurfaceKernel module (this changes the signature of SurfaceKernel constructor).

edoardolombardi:
@andrea-iob as we discussed, if the patches are strict to the related codimension and they don't allow to put mixed higher codimensional element in the same patch (e.g. a 3D surface patch could have 3D surface elements, 2D line elements and 1D vertex elements), personally, I cannot approve the pull request. For the purposes of our applications , mixed-codimensional elements have to be supported at least for one type of patch.

andrea-iob:
This is the first time I hear that the reason why this pull request has not been approved it's because it limits the type of elements that can be added to a patch. This limitation is only introduced in the last commit. I've no problems removing that commit (cb072bd). We can also ditch that commits that change the argument passed to extractEdgeNetwork and extractEnvelope (
7549c34 and 892c81c).

edoardolombardi:
For me it's ok, if the compatibility with mixed-codimension is preserved we can modify as you suggested to go forward with the pull.

edoardolombardi (approved):
I've tried only the compatibility of codes using SurfUnstructured as before, i.e. with mixed elements. No LineKernel & PointKernel embedded in codes and tested outside bitpit.

PR #40, opened 2020-01-29
PABLO: fixed find neighbours methods for root octant

edoardolombardi:
In case of octree composed by root octant only, the neighbours research is not performed.

### DIFF
--- a/src/PABLO/LocalTree.cpp
+++ b/src/PABLO/LocalTree.cpp
@@ -902,6 +902,17 @@ namespace bitpit {
             return;
         }
 
+        // Exit if is the root octant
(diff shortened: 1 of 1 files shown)


andrea-iob:
I've updated the comment and fixed the indentation. If changes are fine I will do the merge.

andrea-iob:
@edoardolombardi Do find neighbour functions support periodic trees? If you have a periodic tree with just the root octant and you ask the neighbours of the root octant, should PABLO return the root octant itself?

edoardolombardi:
Yes, I think it is more correct. I modified the related method.

PR #41, opened 2020-01-30
voloctree: add parallel mapping

edoardolombardi:
Add parallel version of voloctree mesh mapper.


andrea-iob:
Please stash together all the POD/mapper changes after "first version of parallel mapping with update" (maybe also the previous commits can be stashed together). Some commits doesn't compile, so there is not much sense keeping them separate (each commit on the master branch should compile). Plus, it is difficult to review the changes because code that is added in one commit is fixed/changed/deleted some commits later.

We cannot call the mapper MeshMapper, because in bitpit the meshes are called patches. 

Let's put this new MeshMapper in its final place. If needed, we can add a new module (we may call it mapping or something similar).

edoardolombardi:
As required:
- Commits squashed together in two main commits
- Names are now updated coherently to bitpit classes
- Mapper classes placed in the related patch-type modules

src/patchkernel/volume_mapper.cpp
```diff
@@ -0,0 +1,207 @@
...
+: m_referenceMesh(nullptr), m_mappedMesh(nullptr), m_mapper(1)
+{
+
+#if BITPIT_ENABLE_MPI
+    m_communicator = MPI_COMM_NULL;
+    initializeCommunicator(comm);
+    MPI_Comm_size(m_communicator, &m_nProcs);
+    MPI_Comm_rank(m_communicator, &m_rank);
+    m_rank = 0;
+    m_nProcs = 1;
```
andrea-iob: @edoardolombardi It seems strange that the values of m_rank and m_nProcs are overwritten just after being set.

src/patchkernel/volume_mapper.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+}
+
+/**
+ * Destructor of VolumeMapper
+ */
+VolumeMapper::~VolumeMapper()
+{
+}
+
```
andrea-iob: @edoardolombardi No need to use empty destructors, just use "virtual ~VolumeMapper() = default;" in the header.

src/voloctree/voloctree_mapper.hpp
```diff
@@ -32,102 +32,41 @@
 #include <vector>
 #include <unordered_map>
 
+#include "volume_mapper.hpp"
 #include "bitpit_voloctree.hpp"
```
andrea-iob: @edoardolombardi Do not include "bitpit_voloctree.hpp" inside the voloctree module. Include the specific headers you need.

edoardolombardi:
Settled. Thanks.

andrea-iob:
What I meant with "squash together all the POD/mapper changes" was to merge together only the changes related to the mapper keeping all other changes separate. The first of the two merged commits doesn't compile. I've refactored the commits here:

 https://github.com/andrea-iob/bitpit/pull/new/voloctree.add.parallel.mapping

The refactored commits does the following:
 - preparatory changes to PABLO and POD;
 - introduce VolumeMapper;
 - introduce VolOctreeMapper;
 - modify POD module to use VolOctreeMapper;

During the refactoring I've done some changes to the coding style to make it consistent with the coding style of the other files of the same modules. You will also find some commits marked with "xxx" which needs your approval:
 - the new function ParaTree::getInternalOctants is not needed, it can be replaced by the already existing VolOctree::getInternalOctants;
 - use emplace_back and reserve to optimize the generation of OctantIR lists;
 - avoid using all uppercase names for variables;
 - use lowercase name for id;
 - avoid taking references of small data types;
 - no need to define empty destructors;
 - avoid defining default move constructors;
 - avoid using two arguments with almost an identical name (infoAdapt and InfoAdapt);
 - pass output arguments by pointers;
 - make the function PODKernel::_computeMapper return the mapper;
 - reduce the scope of some variable;
 - remove some unused variables.

There is duplicate code in some core functions, in the future it will be worth to invest some time in removing this code duplication (e.g., defining some helper functions).

andrea-iob:
Another aspect that may need some optimization is the fact the the data structure OctantIR is storing a copy of the octant. Would it be reasonable to store only the octants not in the local patch and use pointers to the octants in the local tree?

edoardolombardi:
Thank you for the makeover of the pull. Changes and new commits are ok for me.

Just a little fixing is needed in mapper example 3 & 4: all the function calls "get[Received][Sent]MappedID()" have to be replaced with get[Received][Sent]MappedIds(), coherently with the new names. 

https://github.com/optimad/bitpit/blob/eca5b3abd70c099323e3a2fad760282770ce4b41/examples/voloctree_mapper_example_00003.cpp#L279-L280

Can you add this modifications to the related commit 1bf23b87fabcd495afed9276d704c03509accc91 , please? (I've already a commit for that, let me know if I have to push it and where). 


> There is duplicate code in some core functions, in the future it will be worth to invest some time in removing this code duplication (e.g., defining some helper functions).

I agree with you, two functions of voloctree mapper class are very similar, but it seems to me that they check often different conditions and different communications are done. We can investigate and work on it in future developments.


> Another aspect that may need some optimization is the fact the the data structure OctantIR is storing a copy of the octant. Would it be reasonable to store only the octants not in the local patch and use pointers to the octants in the local tree?

Yes, I agree. I modified one of the structure with copy of OctantIR (map_rank_rec_octantIR) in order to use pointer of already stored OctantIR objects. Again, I've already a commit with this modification, let me know where I can push it or share the code.

PR #47, opened 2020-02-17
PABLO: fix iterators in update ghosts method

edoardolombardi:
Fixed updating of iterators for internal and border of process octants during ghosts recovering


marcocisternino:
Meglio così o meglio il revert di 876a19638204ebbe774931f4e7006da1c7415b47 ?

edoardolombardi:
I think it is better to maintain the current commit, with resize and shrink.

PR #51, opened 2020-03-02
discretization: add classes to generate reconstruction polynomial

andrea-iob:
There are two main classes: "ReconstructionAssembler" which allows to assembly a reconstruction polynomial and "ReconstructionPolynomial" which allows to apply a reconstruction polynomial previously assembled.

@haysam Is still some work pending or can we merge the changes into the master branch?


andrea-iob:
I added some functions to generate limited reconstructions and updated the function that evaluate the stencil to accept the order of the requested stencil.

andrea-iob:
The class ReconstructionPolynomial needs still some work, as it doesn't support stream operators needed for MPI communications.

PR #53, opened 2020-03-04
LA: improve error messages and allow to specify dump format

marcocisternino:


andrea-iob:
I don't want to expose PETSc data structures through the bitpit API if is not strictly necessary. It would be better to define an enum to allow the selection of ASCII or binary formats (see below). 

I don't like the idea to always print messages on system convergence with no option to disable them. Also, information about convergence are accessible though the KSPStatus data structure (see getKSPStatus()). After solving the system, an application using bitpit can inspect convergence information and, if needed, print some messages.

The changes introduced some warnings:
```
warning: comparison with string literal results in unspecified behavior [-Waddress]
   23 | #define PETSCVIEWERASCII        "ascii"
      |                                 ^~~~~~~
/home/andrea/work/customers/optimad/present/20150716_bitpit/code/src/LA/system_solvers_large.cpp:857:23: note: in expansion of macro ‘PETSCVIEWERASCII’
  857 |     if (matrixType == PETSCVIEWERASCII) {
      |                       ^~~~~~~~~~~~~~~~
/usr/lib/hpc/gnu9/openmpi3/petsc/3.12.2/include/petscviewer.h:23:33: warning: comparison with string literal results in unspecified behavior [-Waddress]
   23 | #define PETSCVIEWERASCII        "ascii"
      |                                 ^~~~~~~
/home/andrea/work/customers/optimad/present/20150716_bitpit/code/src/LA/system_solvers_large.cpp:878:20: note: in expansion of macro ‘PETSCVIEWERASCII’
  878 |     if (rhsType == PETSCVIEWERASCII) {
      |                    ^~~~~~~~~~~~~~~~
/usr/lib/hpc/gnu9/openmpi3/petsc/3.12.2/include/petscviewer.h:23:33: warning: comparison with string literal results in unspecified behavior [-Waddress]
   23 | #define PETSCVIEWERASCII        "ascii"
      |                                 ^~~~~~~
/home/andrea/work/customers/optimad/present/20150716_bitpit/code/src/LA/system_solvers_large.cpp:899:25: note: in expansion of macro ‘PETSCVIEWERASCII’
  899 |     if (solutionType == PETSCVIEWERASCII) {
```

In the branch "LA.SystemSolver.dump.convergence.ai" on my fork, I've modified your changes in order to introduce a DumpFormat enum to specify the format of the dump (this also removes the warnings). I've also split the first commit, because it was introducing three different changes. In that branch I've also dropped the patch the introduced the messages. If they are really needed we can put them back, but they should be optional (e.g., display the messages only if a verbose flag is passed to the solve function).

marcocisternino:
No problem. Much better. Let's move messages to the app, collecting most interesting info from petsc in bitpit. Viewer types are more than two, we should enlarge the enum.
I briefly saw your fork, did you mix ASCII and binary ifs for setting the viewer format?
We should put in a method the petsc calls to set a viewer, too duplicated code.

Ottieni Outlook per Android<https://aka.ms/ghei36>

________________________________

marcocisternino:
No problem. Much better. Let's move messages to the app, collecting most interesting info from petsc. Viewer types are more than two, we should enlarge the enum.I briefly saw your fork, did you mix ASCII and binary ifs for setting the viewer format?
We should put in a method the petsc calls to set a viewer, too duplicated code.

PR #54, opened 2020-03-09
patchkernel: fix data types of VTK geometry fields

andrea-iob:
Data type of offset field should match the data type of the connectivity. 

This is the error raised by Paraview 5.8 while opening a VolOctree patch:
```
vtkXMLUnstructuredGridReader (0x5594e13b0b20): Cannot read cell data from Cells. Offsets and connectivity arrays must be the same type.
```

@haysam The change to the IO module is not strictly needed to fix the issue because PatchKernel is not using VTK homogeneous grids, but I think we shouldn't set the type of the offset field to UInt64 unconditionally (it may not match the data type of the connectivity field).

### DIFF
--- a/src/IO/VTK.hpp
+++ b/src/IO/VTK.hpp
@@ -400,12 +400,14 @@ class VTKUnstructuredGrid : public VTK {
             private:
                 VTKElementType          m_type ;                    /**< the type of cells */
                 long                    m_nCells ;                  /**< numer of cells */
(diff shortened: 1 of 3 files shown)


andrea-iob:
The data type used when flushing vertex index was wrong, I fixed it as well.

edoardolombardi:
For me it's ok, but I leave @haysam approving as author of the module.

andrea-iob:
@haysam Can I merge this?

PR #56, opened 2020-03-13
examples: add an example to demonstrate VolOctree mesh adaptation

andrea-iob:
The example creates a three-dimensional octree mesh on the square domain [0,1]x[0,1]. On this domain two fields, a scalar and a vector, are defined.

Mesh is first refined and then coarsened, during each mesh adaption step fields are remapped on the updated mesh.


andrea-iob:
Can I merge this changes to the master branch?

Issue #23, opened 2019-11-14
CMake error: PETSC could not be found


JanssensNick:
Dear,
When using ccmake to install bitpit, I get the following error:

 CMake Error at
 /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137
 (message):
   PETSc could not be found.  Be sure to set PETSC_DIR and PETSC_ARCH.
   (missing: PETSC_INCLUDES PETSC_LIBRARIES PETSC_EXECUTABLE_RUNS)
 Call Stack (most recent call first):
   /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378
 (_FPHSA_FAILURE_MESSAGE)
   external/PETSc/cmake/FindPETSc.cmake:348
 (find_package_handle_standard_args)
   CMakeLists.txt:524 (find_package)

However, I have set all environmental variables (PETSC_DIR, PETSC_ARCH, PETSC_INCLUDES, PETSC_LIBRARIES and PETSC_EXECUTABLE_RUNS). Does anybody know how to solve this?

marcocisternino:
Hello Nick,
we are glad to know about your interest in bitpit.
About your problem, you should give us some more information. I'll try to guess:
- did you install PETSc or you only build it? Please, install it and set PETSC_DIR and PETSC_ARCH according to your installation.
- If you already install PETSc, please, verify that the values you are using in PETSC_DIR and PETSC_ARCH are the one suggested by PETSc after installation and that the path $PETSC_DIR/$PETSC_ARCH exists, a common installation leaves PETSC_ARCH empty.

If none of the guesses above solve the problem and you installed PETSc, please send us you PETSC_DIR and PETSC_ARCH values after installation, the PETSc configure.log and the file make.log (this file is in a subfolder of your build folder and this subfolder has the name of the value used for PETSC_ARCH during your building i.e. YOUR_BUILD_FOLDER/$PETSC_ARCH/lib/petsc/conf/make.log ).

I hope this can help you. Please, let us know.

Thanks,

Marco Cisternino, PhD
marco.cisternino@optimad.it
_______________________________
OPTIMAD Engineering srl
Via Giacinto Collegno 18, Torino, Italia.
+3901119719782
www.optimad.it


________________________________________
Da: Nick Janssens <notifications@github.com>
Inviato: giovedì 14 novembre 2019 12:52
A: optimad/bitpit
Cc: Subscribed
[...]

PR #57, opened 2020-03-14
cmake: rework handling of external dependencies

andrea-iob:
This rework fixes two issues:
 - on some distributions (i.e. OpenSUSE) MPI Fortran library is needed for compiling PETSc;
 - bitpit will no longer shadow PETSC_ARCH and PETSC_DIR environmental variables (see #23).


andrea-iob:
Changes passed all automatic tests and all manual tests I did during deveopment. However, changes to the build system may introduce errors that happen only on some specific configurations/distributions. The more tests the better (were "test" means compile bitpit starting from an empty build directory).

marcocisternino:
I pushed a commit that should fix the env variables shadowing.
I had on this branch the same shadowing problem I have on master.
Now it disappears.
Please, check it if you can.

andrea-iob:
I fixed the order in which external dependencies were added.

PR #59, opened 2020-03-29
bitpit: fixes for some misusage related to references

andrea-iob:
Fixes the following type of misusages of references:
 - small and trivially-copyable types passed by reference;
 - large types passed by value;
 - usage of copies in for-range loops.

### DIFF
--- a/src/IO/DGF.cpp
+++ b/src/IO/DGF.cpp
@@ -107,7 +107,7 @@ return; }
     \param[in] filename dgf file name
 */
 DGFObj::DGFObj(
-    std::string                          filename
+    const std::string                   &filename
 ) {
 
 // ========================================================================== //
@@ -125,7 +125,8 @@ DGFObj::DGFObj(
 // ========================================================================== //
 
 // General info
-dgf_name = utils::string::trim(filename);
+dgf_name = filename;
+dgf_name = utils::string::trim(dgf_name);
 
 // Error flags
 err = 0;
@@ -148,7 +149,7 @@ return; }
     \param[in] mode opening mode ("in": input, "out": output, "app": append mode)
 */
 void DGFObj::open(
-    std::string                          mode
+    const std::string                    &mode
 ) {
 
 // ========================================================================== //
@@ -211,7 +212,7 @@ return; };
     "app": append mode)
 */
 void DGFObj::close(
-    std::string                          mode
+    const std::string                    &mode
 ) {
 
 // ========================================================================== //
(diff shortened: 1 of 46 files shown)


src/containers/piercedStorage.hpp
```diff
@@ -405,7 +405,7 @@ friend class PiercedStorageIterator;
     template<typename T = value_t, typename std::enable_if<PiercedStorage<T, id_t>::has_restore()>::type * = nullptr>
     void restoreField(std::istream &stream, T &object);
 
-    void dumpField(std::ostream &stream, const std::vector<bool>::const_reference &value) const;
+    void dumpField(std::ostream &stream, std::vector<bool>::const_reference value) const;
```
edoardolombardi: Why not just bool instead of vector::const_reference?

src/IO/DGF.cpp
```diff
@@ -107,7 +107,7 @@ return; }
     \param[in] filename dgf file name
 */
 DGFObj::DGFObj(
-    std::string                          filename
+    const std::string                   &filename
```
edoardolombardi: If I well understood the modifications (from copy value to reference) are not related to the title of the commit.

src/containers/piercedStorage.hpp
andrea-iob: A std::vector<bool> may not be a vector of bool objects. To make it space-efficient, every implementation is free to chose any memory layout it likes to store the elements of that array. Hence a reference to an element of a std::vector<bool> may not be a bool.

src/IO/DGF.cpp
andrea-iob: In my view, "misusages" include also the cases where references should have been used but they were not (i.e., all the "*: pass large types by constant reference" commits).

edoardolombardi: The title of the commit says "IO/DGF: pass small and trivially-copyable types by value". am I wrong ?

src/containers/piercedStorage.hpp
edoardolombardi: Sorry, it's not really clear to me; my comment was related to std::vector<bool> element that I suppose you want to pass by value in this commit.

src/IO/DGF.cpp
andrea-iob: I though you meant the title of the pull request. I will fix the commit message.

src/containers/piercedStorage.hpp
andrea-iob: The operator ```[]``` of a ```std::vector<bool>``` doesn't return a ```bool``` but a ```std::vector<bool>::reference```. 

Just to be sure, I will do some tests to check if, after the change, the above template specialization is still used when dumping a PiercedStorage of bools.

andrea-iob:
I fixed the commit message for the IO/DGF changes and checked that the the dump of a PiercedStorage of bools works as expected.

andrea-iob:
Can I merge this changes to the master branch?

PR #67, opened 2020-06-10
Fix voloctree mapper

edoardolombardi:
Fix inverse mapping re-communication


andrea-iob:
Changes to the .gitignore file doesn't belong to this pull request.

Mapping info can be large, can we avoid making the copy? Something along the lines with (not tested):
```
            // Remove old ids and ranks if rank != from local one
            auto infoItr = info.begin();
            while (infoItr != info.end()) {
                int rank = infoItr->ranks[i];
                if (rank != m_mappedPatch->getRank()){
                    infoItr = info.erase(infoItr);
                } else {
                    ++infoItr;
                }
            }
```

Issue #66, opened 2020-06-09
Cmake general: Communications module not disabled in serial compiling


roccoarpa:
On master branch, tip commit 8cee42839d7ac9f6fafdf43e4267db90f4532a4b:

Enabled serial compiling (ENABLE_MPI = OFF) -> Communications module depending on MPI is not disabled by default.

roccoarpa:
Addressed @commit c88124a12f268e697a0f1e8a200c8be06c7c96ad. 
Cross-referencing pull-req #68

PR #68, opened 2020-06-10
cmake: fix generation of the MPI-only modue list

andrea-iob:
The list of MPI-only modules should be generated unconditionally and should not be deleted after its first usage (the variable is needed later, when initializing modules state).


andrea-iob:
Should fix issue #66.

PR #75, opened 2020-07-15
Windows porting (MinGW)

roccoarpa:
I'm opening temporarily this pull request to track down the developments for bitpit's Windows porting through msys2/mingw.
Personally, I've fully tested it on local Win10 machines, and it works. I've just some doubts on the final expected results of some modules examples, but i think it's something related to my lack-of-knowledge.
@edoardolombardi @marcocisternino, as far as you know,  before officially asking the final review/merge to @andrea-iob , do we need more testing on this branch?

Let me know your opinion, just to decide how to proceed.


edoardolombardi:
Even on my machine compilation and testing are ok.

marcocisternino:
Let's merge!
M

Ottieni Outlook per Android<https://aka.ms/ghei36>

________________________________

src/IO/VTK.cpp
```diff
@@ -695,12 +695,21 @@ void VTK::writeData( ){
...
     if (!str.is_open()) {
         throw std::runtime_error("Cannot create file \"" + m_fh.getName() + "\"" + " inside the directory \"" + m_fh.getDirectory() + "\"");
     }
+    //Position the input stream at the beginning
+    str.seekg(0, std::ios::beg);
 
-    { // Write Ascii
+    { // Write Ascii - It may seem disturbing to write ascii on a binary opened
+      // file but genericIO::flushBinary (reinterpret_cast of chars into chars)
+      // should ensure no surprises into the final file.
```
andrea-iob, line 713: "Should ensure no surprises" is not very reassuring. If we are sure this works let's say that in the comment, otherwise it's better to do some more tests to make sure there are no actual surprises.

src/IO/VTK.cpp
```diff
@@ -722,7 +731,7 @@ void VTK::writeData( ){
                 }
 
                 str << std::endl ;
-                genericIO::flushBINARY( str, buffer, length) ;
+                genericIO::flushBINARY( str, buffer, length) ; // this is basically a reinterpret_cast of chars into chars. It should be safe in this context
```
andrea-iob, line 734: Can we remove this comments and the ones below? The previous comment is already stating why we are using a binary write for an ASCII file.

src/IO/VTK.cpp
```diff
@@ -1013,11 +1008,11 @@ void VTK::readData( ){
...
-        str.close();
-        str.clear();
-
-        //Open in binary for read
-        str.open( m_fh.getPath( ), std::ios::in | std::ios::binary);
+        // str.close();
+        // str.clear();
+        //
+        // //Open in binary for read
+        // str.open( m_fh.getPath( ), std::ios::in | std::ios::binary);
```
andrea-iob, line 1016: Please remove these commented lines.

cmake/CMakeLists.txt
```diff
@@ -46,6 +46,45 @@ foreach(MODULE_NAME IN LISTS BITPIT_MODULE_LIST)
...
+            set(STASHPETSC ${ENTRY})
+        endif()
+    unset(_index)
+    endforeach()
+    if(STASHPETSC)
+        list(REMOVE_ITEM BITPIT_EXTERNAL_LIBRARIES_SANITIZED ${STASHPETSC})
+    endif()
+    unset(STASHPETSC)
+endif()
+#------------------------------------------------------------------------------------#
```
andrea-iob, line 84: Indentation and whitespaces are a bit weird. I will fix them before merging.

README_WINDOWS.md
```diff
@@ -1,261 +1,265 @@
-#  bitpit in Microsoft Windows
+#  bitpit on Microsoft Windows
```
andrea-iob, line 1: Please merge this commit with the one that introduces the README file.

src/IO/VTK.cpp
roccoarpa: Ironically, this was the most stressed part of my tests. I've modified the comment

src/IO/VTK.cpp
roccoarpa: done

README_WINDOWS.md
roccoarpa: done

cmake/CMakeLists.txt
roccoarpa: really nothing to do here. Checked the 4ws indentation and it's correct.

src/IO/VTK.cpp
roccoarpa: done

andrea-iob:
There is one more issue I would like to address before merging the branch into the master (I'm not sure if I've already reported it in the review). I think it's better to remove the duplicates from the variable BITPIT_EXTERNAL_LIBRARIES itself, rather than creating a new variable. Also, I'm not sure if the function REMOVE_DUPLICATES works properly in this case. If the list of external libraries is something like "library_1.a libmpi.a library_2.a libmpi.a", removing the duplicates would give us "library_1.a libmpi.a library_2.a", but this will not work if library_2.a depend on libmpi.

roccoarpa:
> There is one more issue I would like to address before merging the branch into the master (I'm not sure if I've already reported it in the review). I think it's better to remove the duplicates from the variable BITPIT_EXTERNAL_LIBRARIES itself, rather than creating a new variable. Also, I'm not sure if the function REMOVE_DUPLICATES works properly in this case. If the list of external libraries is something like "library_1.a libmpi.a library_2.a libmpi.a", removing the duplicates would give us "library_1.a libmpi.a library_2.a", but this will not work if library_2.a depend on libmpi.

Look, I did some further tests and I realized this REMOVE_DUPLICATES story was really unnecessary to land the commit. So nevermind. I will push an updated version shortly. Hope it satisfies your requests.

andrea-iob:
In my opinion, the problem with the PETSc static library will arise also for all other static variables. I think it would be better to remove all static variables from the list of external libraries exported by bitpit. I've added this change in commit c8aea1d. I've also moved the definition of PETSC_ARCH and PETSC_DIR before finding PETSc package (13463d9).

@marcocisternino confirms that the updated branch compiles on Windows. 

If @roccoarpa confirms that the changes are fine we can merge into master.

roccoarpa:
> In my opinion, the problem with the PETSc static library will arise also for all other static variables. I think it would be better to remove all static variables from the list of external libraries exported by bitpit. I've added this change in commit [c8aea1d](https://github.com/optimad/bitpit/commit/c8aea1ded05d0fc3a44bbb84347d80d857183aac). I've also moved the definition of PETSC_ARCH and PETSC_DIR before finding PETSc package ([13463d9](https://github.com/optimad/bitpit/commit/13463d9075356bf1cea8345f5a4edff39de463ac)).
> 
> @marcocisternino confirms that the updated branch compiles on Windows.
> 
> If @roccoarpa confirms that the changes are fine we can merge into master.

From my side all I need is that the full, shared(dll.a or dll) compiled bitpit can be included in a external project who explicitly uses a cmake find_package(BITPIT) command to retrieve it, without arising dependencies conflicts. If your new modifications/commits achieve it, they are fine for me.  Regards.

andrea-iob (approved):

PR #80, opened 2020-08-26
patchkernel: keep track of vertices owners

andrea-iob:
Each vertex is now associated with a processor, that processor is the owner of the vertex. Given the owners of the cells in one-ring neighborhood of a vertex, the owner of the vertex is the processor with the lowest rank. 

Like for the cells, data structures for ghost data exchange are built also for the vertices.


edoardolombardi (approved):

src/patchkernel/patch_kernel_parallel.cpp
```diff
@@ -2947,6 +3234,54 @@ std::vector<int> PatchKernel::getNeighbourRanks()
 	return neighRanks;
 }
 
+/*!
+	Gets a constant reference to the ghost targets needed for data exchange.
+
```
edoardolombardi, line 3239: A comment here can help to understand better what are ghost targets. Something like:

> For each source rank (key of the returned map) the ordered list of the local ids of the ghost vertices is provided. These vertices are involved as targets in a send/receive data exchange. Targets ghost vertices on a receiving rank have the same order of the source vertices on the related sending rank.

Maybe, it can be written more clearly....

src/patchkernel/patch_kernel_parallel.cpp
```diff
@@ -2947,6 +3234,54 @@ std::vector<int> PatchKernel::getNeighbourRanks()
...
+	exchange for the specified rank.
+*/
+const std::vector<long> & PatchKernel::getGhostVertexExchangeTargets(int rank) const
+{
+	return m_ghostVertexExchangeTargets.at(rank);
+}
+
+/*!
+	Gets a constant reference to the ghost sources needed for data exchange.
+
```
edoardolombardi, line 3263: The same as for target vertices. 
Here, we can also add a comment explaining that the sources are internal vertices and not ghost vertices.

> Sources ids are the ids of the local vertices, owned by the current rank, that are involved as sources in data exchange for each receving rank.

src/patchkernel/patch_kernel_parallel.cpp
```diff
@@ -3088,6 +3465,401 @@ void PatchKernel::clearGhostCellOwners()
...
+	// Sort the targets
+	for (auto &entry : m_ghostVertexExchangeTargets) {
+		std::vector<long> &rankTargets = entry.second;
+		std::sort(rankTargets.begin(), rankTargets.end(), VertexPositionLess(*this));
+	}
+
+	//
+	// Identify exchange sources
+	//
+	// Since vertices are shared among elements, data exchange does not involve
```
edoardolombardi, line 3577: This comment is really clear, maybe can be included in some method documentation?

For instance in a getGhostTarget/SourceVertexExchangeInfo()? Or in the documentation of this method?

Does it make sense?

edoardolombardi (approved):

PR #81, opened 2020-08-26
patchkernel: improve manual update

andrea-iob:
After manual modifications to a patch (e.g., addition or deletion of new cells), it is now mandatory to call the "update()" function. This function will take care of updating all patch data structures (adjacencies, ghost exchange information, bounding box, ...).

The same function allows to initialize a partitioned patch: 
 - before adding vertices/cells to the patch, the (empty) patch should be partitioned with the function "PatchKernel::partition(MPI_Comm)";
 - internal and ghost cells should then be added;
 - calling the update function will create the data structures needed for ghost data exchange.  

(Test "test_surfunstructured_parallel_00007" shows how to initialize a partitioned patch.)

This branch include also the changes of pull request #80.


edoardolombardi:
I pushed two fixing commits:
- the assignement during the partition of the rank owner value not set to -1 in case of distributed mesh;
- in the isDistributed method opposite condition check on the owner value used;

andrea-iob:
I've added a function to initialize the adjacencies: "initializeAdjacencies". 

The function "buildAdjacencies" is now deprecated.

edoardolombardi:
@andrea-iob in the temporary pushed commit d718e68 I added an output to a test useful to check a possible bug in the interfaces update.
The interface ids of some cells seem wrong (unintialized values?) after the second partitioning of the test. 
The same happens when the interfaces are updated after some add/delete operations on cells. In my test the problem was related with ghost cells.

src/patchkernel/patch_kernel_parallel.cpp
```diff
@@ -3460,18 +3470,64 @@ void PatchKernel::unsetGhostCellOwner(int id)
...
 	m_ghostCellOwners.clear();
+
+	setPartitioningInfoDirty(true);
+}
+
+/*!
+	Checks if the information related to the partitioning are dirty.
+
+	\result Returns true if the information related to the partitioning are
+	dirty, false otherwise.
```
edoardolombardi, line 3555: Maybe, we can inform that is a collelctive function. The same for other collective get-functions.

src/patchkernel/patch_kernel.cpp
```diff
@@ -4880,6 +4893,38 @@ void PatchKernel::setAdjacenciesBuildStrategy(AdjacenciesBuildStrategy status)
 	m_adjacenciesBuildStrategy = status;
 }
 
+/*!
+	Checks if the adjacencies are dirty.
+
```
edoardolombardi, line 4835: Here, missing parameter global doc

src/patchkernel/patch_kernel.cpp
```diff
@@ -5237,32 +5268,187 @@ void PatchKernel::setInterfacesBuildStrategy(InterfacesBuildStrategy status)
 	m_interfacesBuildStrategy = status;
 }
 
+/*!
+	Checks if the interfaces are dirty.
+
```
edoardolombardi, line 5197: Missing documentation of parameter global

edoardolombardi (approved):

PR #82, opened 2020-08-26
patchkernel: initialize partitioning in the constructor

andrea-iob:
Setting the communicator can now be done only when the patch is partitioned.


andrea-iob:
After some back and forth, I think the best solution is to initialize partitioning in the constructor. I've updated the constructors of the patches that support partitioning to accept the extra arguments needed to initialize the partitioning. 

For now, it is still possible to initialize a serial patch and the partition the patch later, but this will not be possible in bitpit v1.8.

@edoardolombardi @marcocisternino Do you see any problem in this approach?

andrea-iob:
@edoardolombardi @marcocisternino I had to make some changes in PABLO to allow moving all the octants on the first process during in the constructor. Please check if the changes look fine. I think they may be still some issues when the partitioning moves all the octants on a rank that is not the first one. This is something that should not happen with the current code, but, if this is something we want to support, some additional work may be needed.

src/patchkernel/patch_kernel_parallel.cpp
```diff
@@ -1520,17 +1552,7 @@ void PatchKernel::partitioningCleanup()
 */
 bool PatchKernel::isPartitioned() const
```
edoardolombardi, line 1553: Should this wrapper function isPartitioned used at the place of isCommunicatorSet function? In this case is it better to use it everywhere even in PatchKernel and its derived classes to avoid the procreation of re-coding around, in case of change in the wrapper function?

edoardolombardi:
- https://github.com/optimad/bitpit/pull/82#issuecomment-753887911  For me it's ok. One limitation that I see: in case of restore of a patch the user has to know a-priori if the restoring patch has been constructed as parallel or serial. This can be a problem? 

- https://github.com/optimad/bitpit/pull/82#issuecomment-753900220 Even for PABLO, some workarounds (e.g. partitioning with an ad-hoc weights data structure) are already allowed to reach the same result, right? I think we can postpone the implementation of such a partitioning function to future needs.

edoardolombardi (approved):
@andrea-iob just check my previous comments before merging, please.

src/patchkernel/patch_kernel_parallel.cpp
andrea-iob: Done.

Please note that some functions that handle the communicator may be simplified when the deprecated partitioning functions (the ones that take the communicator as argument) are be removed.

andrea-iob:
>     * [#82 (comment)](https://github.com/optimad/bitpit/pull/82#issuecomment-753887911)  For me it's ok. One limitation that I see: in case of restore of a patch the user has to know a-priori if the restoring patch has been constructed as parallel or serial. This can be a problem?

The same limitation exists in the current master: the restore function has to be called by the same number of processes that have called the dump function, this means that for restoring a patch it is necessary to know how it was dumped. To overcome this limitation, dump and restore should be handled by the PatchManager. Before dumping the patch, the manager could write some additional information that can later be used to identify how the patch should be restored. This will also allow to restore a patch without knowing the type of the patch (e.g., VolOctree, SurFunstructured, ...).

andrea-iob:
I fixed the evaluation of PABLO maximum depth when the tree is empty. Also, it seems that the maximum depth was not updated after coarsening. @marcocisternino @edoardolombardi Could you please check if the changes are fine ("PABLO: fix evaluation of maximum depth when the tree is emtpy" 84f4096  and "PABLO: update maximum depth after coarsening the tree" fb5e42a)?

marcocisternino:
Max local depth fixing seems good to me. I tested the branch with a complex mesh using CFD solver. No problem found.

Issue #65, opened 2020-05-31
Simplify DiscretizationStencil class


andrea-iob:
Current implementation of DiscretizationStencil class allows to store different set of stencils. Each stencil is stored in a different "bucket". This feature was used to store in one single data structure the stencils of the different interfaces of a cell.

Since some months, we no longer need DiscretizationStencil class being able to store different set of stencils. Can we remove this feature? @edoardolombardi @roccoarpa Are you relying on stencil buckets in your applications?

roccoarpa:
@andrea-iob: I'm pretty sure we are not using them. But @edoardolombardi please correct me if I'm wrong.

edoardolombardi:
No, we're not using them.

andrea-iob:
Stencil "bucket" feature has been removed.

PR #90, opened 2020-09-26
discretization: remove stencil buckets

andrea-iob:
Stencils are now vector-like containers, it is no more possible to group items in buckets.


haysam (approved):

andrea-iob:
See issue #65.

marcocisternino (approved):

PR #92, opened 2020-10-01
PABLO: optimize find-neighbours functions

andrea-iob:
With these changes, find-neighbours functions are about 20% faster.

### DIFF
--- a/src/PABLO/LocalTree.cpp
+++ b/src/PABLO/LocalTree.cpp
@@ -962,7 +962,7 @@ namespace bitpit {
         neighbours.clear();
 
         // Default if iface is nface<iface<0
-        if (iface > m_treeConstants->nFaces){
+        if (iface >= m_treeConstants->nFaces){
             return;
         }
 
--- a/src/PABLO/LocalTree.cpp
+++ b/src/PABLO/LocalTree.cpp
@@ -3227,7 +3227,7 @@ namespace bitpit {
      */
     uint32_t
     LocalTree::findMorton(uint64_t Morton) const {
-        return _findMorton(Morton, m_octants);
+        return findMorton(Morton, m_octants);
     };
 
     // =================================================================================== //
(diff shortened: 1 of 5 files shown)


andrea-iob:
I've rebased the branch on the current master.

marcocisternino (approved):
Now

marcocisternino (approved):

src/PABLO/LocalTree.cpp
```diff
@@ -3222,53 +2992,156 @@ namespace bitpit {
...
+    LocalTree::findMortonLowerBound(uint64_t targetMorton, const octvector &octants, uint32_t *lowerBoundIdx, uint64_t *lowerBoundMorton) const {
+
+        uint32_t nOctants = octants.size();
+
+        uint32_t lowIndex  = 0;
+        uint32_t highIndex = nOctants;
+        uint32_t midIndex  = nOctants;
+        uint64_t midMorton = PABLO::INVALID_MORTON;
+        while (lowIndex < highIndex) {
+            midIndex  = lowIndex + (highIndex - lowIndex) / 2;
```
edoardolombardi, line 3067: @andrea-iob if I well understand here you're placing always in the worst case. The information on the face (or other entity) through you're searching for the neighbour can be used to fix the portion of the octants container containing the target morton. For mesh with large number of cells I think that the saved time can become considerable.

andrea-iob: The first step of the binary search is always halfway from the beginning and the end of the container. Previous code was using information on the current octant to adjust the first step of the binary search. However, adjusting the first step of the binary search requires the evaluation of one Morton number and, in terms of computational costs, this is equivalent to one step of binary search. Also, it seems to me, that there were cases where the guess for the initial step provided by the old code was just marginally better than starting halfway in the vector. Therefore, I preferred clean code over a small performance gain. If you think there are cases were the performance gains are noticeable we can re-implement the guess for the initial step later.

edoardolombardi: I've noticed in the old code some performance worsening for large meshes. However, the old version was badly coded, so maybe it was caused by bad implementation. 
Concerning the first Morton evaluation, maybe you don't need it, but the linear index of the current octant could be enough.
Anyway, we have just to remeber, in case of future slowdowns in applications, to check the effect of this initial guess in neighbour research.

andrea-iob:
I've cleaned up the commits.

PR #96, opened 2020-10-02
patchkernel: simplify function that get vertex coordinates

andrea-iob:
There is no need to use a ProxyVector to store vertex coordinates, just use plain pointers.


haysam (approved):
The last commit message states that a check is implemented if the type of element is supported. Is this true?

andrea-iob:
> The last commit message states that a check is implemented if the type of element is supported. Is this true?

Fixed. 

I've replaced the warning with an exception, because if some elements of the segmentation are not supported the levelset cannot work properly.

PR #103, opened 2020-10-22
PABLO: rework partitioning 

andrea-iob:
There where two distinct privateLoadBalance implementations (one for partitioning the tree in presence of used defined data and one for partitioning the tree without user defined data). I've simplified and merged the two implementations together. The main simplification is that partitioning information are now obtained from the send/receive ranges.

This rework fixes a segfault in the PABLO_bubbles_3D example.


edoardolombardi (approved):
The bubbles examples with this implementation have no more segfault.

marcocisternino (approved):

PR #104, opened 2020-10-23
PABLO: store Morton number instead of octant's coordinates

andrea-iob:
Instead of storing octant's coordinates, we can store the Morton number. This allows to reduce PABLO memory footprint and speeds up neighbour search.

(This branch starts from PABLO.optimize.find.neighbours.)


edoardolombardi (approved):
If this modifications in general increase the performances, for me the implementation it's ok. I checked out tests and bubbles examples.

marcocisternino (approved):

PR #106, opened 2020-10-29
cmake: add functions to define deprecated/experimental features

andrea-iob:
Experimental feature are disabled by default, whereas deprecated features are enabled by default.

Cmake variables related to experimental and deprecated functions are advanced variables and they will only be visible if the variables BITPIT_SHOW_EXPERIMENTAL or BITPIT_SHOW_DEPRECATED are set to ON.

For each deprecated/experimental feature a corresponding preprocessor macro will be defined. Given the feature FEATURE in the module MODULE, the corresponding preprocessor macro will be called BITPIT_MODULE_FEATURE. The value of the macro will be set to 1 if the feature is enable or it will be set to 0 if the feature is disabled.

Deprecated and experimental should be defined in the section "# Features" of the main CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,62 @@ function(addModuleIncludeDirectories MODULE_NAME)
     include_directories("${PROJECT_SOURCE_DIR}/src/${MODULE_NAME}")
 endfunction()
 
+# Define an experimental feature
+#
+# Experimental features are disabled by default.
+function(defineExperimentalFeature FEATURE_NAME FEATURE_DESCRIPTION)
+    defineFeature("EXPERIMENTAL_${FEATURE_NAME}" ${FEATURE_DESCRIPTION} 0)
+endfunction()
+
+# Define a module experimental feature
+#
+# Experimental features are disabled by default.
+function(defineExperimentalModuleFeature MODULE_NAME FEATURE_NAME FEATURE_DESCRIPTION)
+    string(TOUPPER ${MODULE_NAME} UPPERCASE_MODULE_NAME)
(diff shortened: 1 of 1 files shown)


alessandro-alaia (commented):
Nice 👍! I really like the idea of defining experimental/deprecated on a module/feature basis.
One problem I foresee: what if a deprecated/experimental feature depends on another deprecated/experimental feature. Is there a way to automatically deal with these "internal" dependencies? At the moment I don't think this should be a problem. In the next future, it might become cumbersome due to delay in deploying/removing of experimental/deprecated feature while other experimental features are being developed.

What I find a little confusing is the machinery of "first-showing-then-(de)activating". If I understand correctly you first need to enable "SHOW_deprecated/experimental" features/modules and then you can activate/de-activate such features. Any reason why activation/deactivation cannot be always exposed?

andrea-iob:
I didn't think about dependencies yet. We have some code to track the dependencies among modules, maybe we can borrow some of that code and use it to track feature dependencies.

I'm using the "SHOW_deprecated/experimental" variables as a way to make visible only deprecated or experimental features. There is no difference between variables created by defineDeprecatedModuleFeature and variable created by defineExperimentalModuleFeature (except for their initial state). Maybe be can simply merge the two functions together (defineModuleFeature) and remove the "SHOW_deprecated/experimental" variables.

andrea-iob:
I did some changes:
 - cmake variables "SHOW_DEPRECATED/EXPERIMENTAL" are gone;
 - the keyword DEPRECATED/EXPERIMENTAL is automatically added to both cmake variable and preprocess macro (e.g., the experimental feature FEATURE in the module MODULE will be associated with the cmake variable BITPIT_MODULE_EXPERIMENTAL_FEATURE).

alessandro-alaia (approved):
Seems pretty clear to me and I don't see any potential issue. Merge is approved from my side.

PR #112, opened 2020-11-16
cmake: applications that use bitpit should not relay on information gathered at configuration time

andrea-iob:
@alessandro-alaia noticed that BITPITConfig.cmake adds the external dependencies found during bitpit configuration to BITPIT_LIBRARIES, but, at the same time, it is also using the cmake find infrastructure to find the same dependencies. 

Is there any specific reason why we need both? 

In this pull request I'm proposing to remove the calls to the cmake find functions and relay entirely on the libraries found during bitpit configuration.


edoardolombardi:
> Is there any specific reason why we need both?

I suppose no, there is no reason to have both finding dependencies.

Your proposal fix the double search, but I think that is not a good solution to hard set the path used during bitpit compilation in external dependencies. 

A solution is to maintain the auto-find dependencies of bitpit (that I like) by keeping the find_package calls and appending the correct exposed variables to bitpit dependencies. Indeed, currently they are found but it seems to me that they are never used in bitpit configuration.
Obviously, we have to provide all the Find<package>.cmake files not provided by packages installation, but I think that this logic is already adopted (i.e. for cblas, lapacke, petsc). 

Just as an example we can modify the find dependencies loop in BITPITConfig.cmake: 

```
# Find external packages
list(REVERSE _EXTERNAL_DEPENDENCIES)
foreach(_DEPENDENCY ${_EXTERNAL_DEPENDENCIES})
    find_package("${_DEPENDENCY}" REQUIRED)
    if (${_DEPENDENCY} STREQUAL "LibXml2")
        list(APPEND BITPIT_LIBRARIES ${LIBXML2_LIBRARY})
        list(APPEND BITPIT_INCLUDE_DIRS ${LIBXML2_INCLUDE_DIR})    
    elseif (${_DEPENDENCY} STREQUAL "PETSc")
        list(APPEND BITPIT_LIBRARIES ${PETSC_LIBRARIES})
        list(APPEND BITPIT_INCLUDE_DIRS ${PETSC_INCLUDES})    
        list(APPEND BITPIT_INCLUDE_DIRS ${PETSC_INCLUDE_CONF})    
        list(APPEND BITPIT_INCLUDE_DIRS ${PETSC_INCLUDE_DIR})    
    else()    
        list(APPEND BITPIT_LIBRARIES ${${_DEPENDENCY}_LIBRARIES})
        list(APPEND BITPIT_INCLUDE_DIRS ${${_DEPENDENCY}_INCLUDES})    
    endif()
endforeach()
```

(One if scope it should be coded for each package without default ${_DEPENDENCY}_LIBRARIES and ${_DEPENDENCY}_INCLUDES names)

In this way the last two hard sets are useless and they can be erased:

```
# Add external libraries
list(APPEND BITPIT_LIBRARIES "@BITPIT_SHARED_EXTERNAL_LIBRARIES@")

# Add external inclued paths
list(APPEND BITPIT_INCLUDE_DIRS "@BITPIT_EXTERNAL_INCLUDE_DIRS@")
```

andrea-iob:
I've implemented your idea. 

Tests were performed only on Linux.

edoardolombardi (approved):

andrea-iob:
Can someone tests these changes on Windows?

roccoarpa:
> Can someone tests these changes on Windows?

It works.
Tested with pull rebased on actual master tip commit : https://github.com/optimad/bitpit/commit/53b82749aa03676c3f0912a333ab1fb9454bdaca

roccoarpa (approved):

andrea-iob:
Thanks for the tests. I've rebased the branch on the current master. Once the automatic tests finish I will do the merge.

PR #114, opened 2020-11-20
volcartesian: add functions to get vertex/centroid positions along a direction

andrea-iob:
Add functions to get vertex coordinates and cell centroid positions along a specified direction.

### DIFF
--- a/src/volcartesian/volcartesian.cpp
+++ b/src/volcartesian/volcartesian.cpp
@@ -46,7 +46,7 @@ namespace bitpit {
 	to reduce its memory footprint, it will also limit the functionalites
 	available when operating in light memory mode. Features that are
 	available in light memory mode are:
-	- evaluation of vertices/cell/interface topological properties;
+	- evaluation of vertex/cell/interface topological properties;
 	- evaluation of cell/interface geometrical properties;
 	- evaluation of cell neighbours.
 	Features that are NOT available in light memory mode are:
@@ -872,7 +872,7 @@ std::array<int, 3> VolCartesian::getInterfaceCountDirection(int direction)
 /*!
 	Creates the interfaces normal to the given direction.
 
-	\param direction the method will creat the interfaces normal to this
+	\param direction the method will create the interfaces normal to this
 	                 direction
 */
 void VolCartesian::addInterfacesDirection(int direction)
--- a/src/volcartesian/volcartesian.cpp
+++ b/src/volcartesian/volcartesian.cpp
@@ -481,6 +481,17 @@ long VolCartesian::getCellCount() const
 }
 
 /*!
+	Gets the number of cells along the specified direction.
+
+	\param direction is the direction along which cell count is requested
+	\return The number of cells along the specified.
+*/
+int VolCartesian::getCellCount(int direction) const
+{
+	return m_nCells1D[direction];
+}
+
+/*!
 	Gets the element type for the cell with the specified id.
 
 	\param id is the id of the requested cell
--- a/src/volcartesian/volcartesian.cpp
+++ b/src/volcartesian/volcartesian.cpp
@@ -2024,4 +2024,15 @@ std::array<double, 3> VolCartesian::evalCellCentroid(long id) const
 	return centroid;
 }
 
+/*!
+	Get cell centroids along the specified direction.
+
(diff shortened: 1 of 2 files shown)


alessandro-alaia:
Maybe adding a function which returns the nr. of mesh vertices/cells along a given cartesian direction could be usefull... Something like:
int getCellCount( int i /*coordinate direction x: 0, y: 1, ...*/ ){ return m_nCells1D[i]; }
(and similarly for mesh vertices).

Please fix/integrate the doxygen documentation (avoid **generic comments which simply re-state the name of the member function**).
Add unit tests.

andrea-iob:
> Maybe adding a function which returns the nr. of mesh vertices/cells along a given cartesian direction could be usefull... Something like:
> int getCellCount( int i /_coordinate direction x: 0, y: 1, ..._/ ){ return m_nCells1D[i]; }
> (and similarly for mesh vertices).

Done
 
> Any particular reason why the method evalVertexCoords is privite as opposed to evalCellCentroid? Is not the user supposed to inquiry the mesh to get the coordinate of a vertex from its cartesian/rectilinear ID? Make the method public, otherwise there is no way to access vertex coordinates from Cartesian IDs (i,j,k).

PatchKernel stores a Vertex object for each vertex, vertex coordinates are accessible through that data structure (patch.getVertex(id).getCoords()). However, for the Cartesian mesh, it make sense to declare the function evalVertexCoords public in order to have a faster access to that information. Moreover, this allows to access vertex coordinates also when operating in light memory mode (when the mesh is in light memory mode, Vertex data structures are not built and hence vertex coordinates would not be accessible).

> Please fix/integrate the doxygen documentation (avoid **generic comments which simply re-state the name of the member function**).

An example on what you mean with integrate would be useful, because the added functions are quite simple and literally return the vertex/centroids coordinates along the requested direction. I can add a comment stating that 0 is direction x, 1 direction y, ... Other that that I don't have much else to say. 

> Add unit tests.

Is it better to add a different test for each function or is it fine to have a single test covering more than one function?

alessandro-alaia:
> An example on what you mean with integrate would be useful, because the added functions are quite simple and literally return the vertex/centroids coordinates along the requested direction

I agree that those functions are quite self-explainatory and do not need to expand the doxygen documentation.

A example of **lack of documentation** is the switchMemoryMode member function, which is documented as
" Switch to the specified memory mode." (which simply state the function name).
What does "memory mode" means **exactly**? What happens to the behavior of the class, how is it modified? Which data will be computed on-the-fly (vs. stored within the class) when switching from one mode to the other? Which functions will be affected? What happens when I switch memory mode? Is the class in good state (ready-to-go), or should the user call the method "update"? If so, what happens if I don't call "update" (does the class throws a exception, should the user expect undefined behavior, nothing happens and the class behaves like before)? How will the memory mode affect the memory usage vs. the CPU usage? Those information are needed so that the user can make informed decisions on how to use the class without having to read the entire implementation and try to figure out **your** design choices.

> Is it better to add a different test for each function or is it fine to have a single test covering more than one function?

I think it should be fine to cluster simple tests all together in the same test.cpp file. Imho, something like this should work robustly:
int test_suite()
{
  int nerrors = test_constructors()
  //break, if (for instance) the test on default constructor fails because all the other tests will automatically fail.
  nerrors += test_operators()
  //break if one operator did not pass the test and some of the subsequent tests make use of that particular operator (they will automatically fail)
  nerrors += test_getters()
  //and so on
  nerrors += test_setters()
  //any other test involving more advanced feature...

  return nerrors;
}
A alternative could be the use of boost test library.

andrea-iob:
I've added an initial implementation of the unit tests in pull request #115.

Improving the documentation is on the TODO list (this applies to the documentation of the whole bitpit librariy). Some of your examples of an incomplete documentation applies to the whole PatchKernel class rather being specific for the VolCartesian patch. As a temporary solution, I can add some brief documentation specific for the VolCartesian class. In the future we will need to harmonize the documentation of all the patches (maybe adding some examples to explain how the patches can be used).

If the added methods are fine with you, we can proceed and merge this pull request. The documentation will be updated in another pull request.

alessandro-alaia (approved):

alessandro-alaia:
All the requested requested feature have been added. No particular concerns on my side about changes introduced in the last commits

PR #115, opened 2020-11-24
test: add unit tests infrastructure

andrea-iob:
This is a first implementation of the unit tests for the VolCartesian patch. 

Tests are distributed among 5 suites:
 - unit_test_volcartesian_constructors: this suite test the constructors;
 - unit_test_volcartesian_general: this suite test methods that works on the patch as a whole;
 - unit_test_volcartesian_vertices: this suite test vertex-related methods;
 - unit_test_volcartesian_cells: this suite test cell-related methods;
 - unit_test_volcartesian_interfaces: this suite test interface-related methods.

(In some cases, the distinction between these 5 categories is a bit fuzzy, but I think the same applies for any other classification we may come up with. I chose these categories, because I liked the idea to have the tests grouped by some macroscopic features).

Inside each suite, test are grouped in two subtests: 
 - subtest_001: tests for patch in normal memory mode; 
 - subtest_002: tests for patch in normal light mode.

Although we may decide to remove the light memory mode feature (at the present time there may be no applications that use that future), I added unit tests also for that mode just for the sake of completeness.

Unit tests cover all public functions defined in VolCartesian header. Functions defined in the base class Patchkernel and not overloaded inside VolCartesian are not yet included in the tests.

I'm not using any special framework for the tests. If we agree on the usage of boosts_test I will port the tests to this framework (or to any other framework we may decide to use, e.g., Google Test, CppUnit, ...).

Test lacks some output messages explaining which test is currently running, I will add these messages once we define which framework use for handling unit tests.


andrea-iob:
This pull request contains also some fixes/clean-ups I made while working on the tests.

alessandro-alaia:
I like this proposal. The lack of messages might become a issue later when the nr. of tests will grow larger, but for now it I don't see a big problem.
Porting those tests in Boost testing library (if we decide to go that way) should also be easy.
Green flag from my side. Thanks! 👍

andrea-iob:
Unit tests now use Boost.Test library and there is an explicit distinction between integration tests (the old tests) and unit tests. The two test categories live now in two different directories. To keep the old behaviour, integration tests are on by default, whereas unit tests needs to be explicitly enabled (cmake variable ENABLE_UNIT_TESTS). When unit tests are enabled, Boost.Test library becomes a required dependency. 

A serial test suite can be defined using this structure:
```
#define BITPIT_UNIT_TEST_SUITE_NAME unit_test_name
#include "helpers/unit_test.hpp"

BOOST_AUTO_TEST_SUITE(suite_name)

BOOST_FIXTURE_TEST_CASE(test_name, Fixture)
{
    BITPIT_UNIT_TEST_DISPLAY_NAME(log::cout());
    ... The actual test ...
}

... Other tests ...

BOOST_AUTO_TEST_SUITE_END()
```

A parallel test suite can be defined using this structure (only the name of the included file changes with respect to a serial test):
```
#define BITPIT_UNIT_TEST_SUITE_NAME unit_test_name
#include "helpers/unit_test_parallel.hpp"

BOOST_AUTO_TEST_SUITE(suite_name)

BOOST_FIXTURE_TEST_CASE(test_name, Fixture)
{
    BITPIT_UNIT_TEST_DISPLAY_NAME(log::cout());
    ... The actual test ...
}

... Other tests ...

BOOST_AUTO_TEST_SUITE_END()
```

I've not yet tested how Boost.Test handles MPI tests, so I'm not sure if parallel unit tests works properly (VolCartesian module doesn't have parallel tests).

andrea-iob:
Next week I'm going to merge this pull request.

PR #118, opened 2020-12-03
cmake: fix order of exported bitpit libraries list

andrea-iob:
Libraries were previously added in the wrong order.


roccoarpa:
_Just a quick note_: @andrea-iob, @marcocisternino, @alessandro-alaia. This pull solves also a compilation bug in **Win-mingw static bitpit** exported libraries, when included by another projects with findPackage. This notification integrates my poor windows review at https://github.com/optimad/bitpit/pull/112, already merged on master. Sorry guys, my bad, I tested only the dynamic lib installation. With this fix, the static part works too.

andrea-iob:
Thanks for testing the changes on Windows. I've added you to the reviewers, feel free to add your review if you think this changes are fine.

Also in my case, the problem was related to the compilation of an external program that is using bitpit compiled as a static library.

roccoarpa (approved):

andrea-iob:
With @marcocisternino we found another place where we were using the external libraries in the wrong order. We also added a workaround to allow compiling bitpit with Intel compiler (I added the fix for the Intel compilers into this pull request, because the change was needed in order to complete the testing).

andrea-iob:
@marcocisternino I've updated the pull request: static libraries are now excluded from the list of dependencies only on Windows.

marcocisternino (approved):

PR #122, opened 2020-12-16
VTK::readDataHeader fix disabling of unused fields

roccoarpa:
changed list type for unusedfields tracking because of moving reference issue after possible push_back on m_data std::vector.


andrea-iob (approved):
Changes look fine to me. I've only slightly changed the commit message.

roccoarpa:
👌

PR #124, opened 2020-12-24
levelset: simplify update

andrea-iob:
There is now a single function that handles both mesh adaption and mesh partitioning.


andrea-iob:
@haysam Can you take a look at this changes?

marcocisternino (approved):

PR #126, opened 2020-12-29
cmake: add support for coverage analysis

andrea-iob:
There is now a new build type ("Coverage") that allows to generate coverage analysis with "make clean && make coverage". At first, tests will be used to gather coverage data, then gcovr or lcov can be used to generate a coverage report. it is possible to select which reports to generate (gcovr, lcov or both) using two cmake variables (BITPIT_COVERAGE_GCOVR_REPORTS, BITPIT_COVERAGE_LCOV_REPORTS).

Coverage analysis can be run only when the code is compiled with gcc.

gcovr report:
gcovr_report

lcov report:
lcov_report


andrea-iob:
Next week I'm going to merge this pull request.

PR #128, opened 2021-01-07
patchkernel: rework function that adjust cell orientation

andrea-iob:
The new implementation properly takes into account border cells, also it doesn't require the interfaces.


marcocisternino (approved):
Travis issues should be cured. Normals are correctly flipped in my tests.

PR #130, opened 2021-01-20
Allow to retrieve extended version information at runtime

andrea-iob:
I've added a function (```bitpit::getVersion()```) that allows to retrieve version information at runtime. The version string contains the bitpit version plus additional information about git branch/commit. For example the version string may be like the following:
```
1.8.0-bitpit.add.git.info.to.version-de4aebf84-dirty
```
where:
 - ```1.8.0``` is the "standard" bitpit version
 -  ```add.git.info.to.version``` is the current branch
 - ```de4aebf84``` is the current commit
 - ```dirty``` is a (configurable) label that tells if the current source has uncommitted changes.

I haven't found a way to automatically re-run cmake (and thus update the version) if there are uncommitted changes. The "dirty" label will be reliably added only if, after changing the code, cmake is run manually.

Git information will be added only if the current HEAD is not associated with a tag in the form "bitpit-X.Y.Z" where "X.Y.Z" should match the version number. In this way git information will be shown only for development builds.


roccoarpa (approved):
Fine for me.

marcocisternino (approved):

PR #138, opened 2021-01-29
levelset: fix evaluation of segment information for points that lie on the segmentation

andrea-iob:
Due to numerical errors, points that lie on the surface may have a distance slightly grater than zero. Therefore, a small tolerance is now used when checking if a point lies on the segmentation.

Moreover, if a point lies on the segmentation, we assume that it is on the outer side of the surface, this means that the distance is a positive zero and that the gradient is set equal to the surface normal.


andrea-iob:
I've updated the branch, when the point lies on the surface we no longer assume it's on the outer side of the surface, but we evaluate the proper sign.

haysam (approved):
I think it is ok and I believe we can omit the check at line 273.

PR #139, opened 2021-02-02
patchkernel: make the MPI communicator a mandatory argument for the constructors

andrea-iob:
Some MPI implementations define the communicator data type as an int. This creates ambiguities in constructor resolution that can
be resolved removing the constructors without the communicator argument.


roccoarpa:
For @edoardolombardi @marcocisternino I've already tested it in MinGW-Win and with mimmo. @andrea-iob It works.

roccoarpa (approved):

PR #145, opened 2021-02-21
cmake: do not enable LTO on MinGW builds automatically

andrea-iob:
LTO doens't work on MinGW.


CMakeLists.txt
```diff
@@ -228,10 +228,14 @@ if(DEFINED LTO_STRATEGY)
...
-                check_ipo_supported(RESULT LTO_SUPPORTED)
-                if (${LTO_SUPPORTED} AND "${CMAKE_BUILD_TYPE}" STREQUAL "Release")
-                    set(ENABLE_LTO TRUE)
+            if (NOT MSYS AND NOT MINGW)
+                if (BUILD_SHARED_LIBS)
+                    check_ipo_supported(RESULT LTO_SUPPORTED)
+                    if (${LTO_SUPPORTED} AND "${CMAKE_BUILD_TYPE}" STREQUAL "Release")
+                        set(ENABLE_LTO TRUE)
+                    else()
+                        set(ENABLE_LTO FALSE)
```
roccoarpa, line 237: Disabling LTO in mingw is quite hidden to the final User here, LTO_STRATEGY remains untouched, leaving the impression it is effectively used. In my opinion, it is better to force LTO_STRATEGY to disabled in MINGW case e raise a -- message(WARNING "LTO is forced to Disabled in MSYS2/MinGW ") -- just to make clear the concept and to make it simpler to re-enable LTO in the future (once the issue will be faced).

roccoarpa (changes requested):

CMakeLists.txt
andrea-iob: The default strategy for LTO is auto and, after these change, it will disable LTO on MinGW. I can update the description of the LTO_STRATEGY in something like "Choose the Link Time Optimization (LTO) strategy, options are: Auto (optimiziation is enabled only in release build and only for configurations that have been tested) Enabled Disabled.".

The fact the LTO doesn't work on MinGW doesn't seem related to bitpit. In my understanding it's a problem/misconfiguration of MinGW. So, if someone wants to enable LTO on MinGW, is free to do so. We can add a warning, stating that MinGW LTO builds will probably don't work.

roccoarpa: Don't matter how, what's important for me is that the User is notified of the situation. A warning raise should  be enough. I agree to pinpoint an issue to investigate the problem/misconfiguration in MinGW.

andrea-iob:
I've updated the patch. Since it seems that making LTO work on MinGW requires changes to bitpit, a cmake error is now issued when trying to enable LTO on MinGW.

roccoarpa (approved):

PR #146, opened 2021-02-21
Fixes and optimization to the evaluation of levelset

andrea-iob:
These change fixes the selection of surface segments when multiple segments are at the same distance from a cell and introduce some miscellaneous optimizations to the evaluation of the levelset.


edoardolombardi (approved):
👍

andrea-iob:
@marcocisternino @edoardolombardi I've updated sign evaluation, it now uses the pseudo-normal. 

@haysam It seems to me that function convertBarycentricToFlagPolygon was giving incorrect results when a point was on an edge, does the fix 54cc90e seems right to you?

The algorithm used to evaluate the pseudo-normal depends on the location of the point within a triangle:
 - if the point coincides with a vertex, the pseudo-normal is evaluated as the unlimited normal of the vertex;
 - if the point lies on an edge, the pseudo-normal is the average of the normals of the two segments sharing the edge;
 - if the point is inside the segment, the pseudo-normal is evaluated as the normal of the segment.

edoardolombardi (approved):

edoardolombardi:
👍  New changes passed my tests.

PR #149, opened 2021-03-04
Fix useBITPIT.cmake file for external projects inclusion

roccoarpa:
Using new call "include_directories(SYSTEM ${BITPIT_INCLUDE_DIRS})", which means to use a -isystem <include_paths> instead of -I<include_paths> for compiler.  This avoids warnings generated by bitpit own and third-party includes to propagate into external projects embedding it.  The end-User should be able to manage only its own project warnings, still using USE_BITPIT_FILE inclusion, eventually.

The workaround is compliant with gcc and clang.
Tested on Linux and MSYS2/MINGW windows 

**STAND-BY**


andrea-iob:
A better solution would be to fix the warnings. What are the warnings you are seeing? Maybe they are fixable.

roccoarpa:
> A better solution would be to fix the warnings. What are the warnings you are seeing? Maybe they are fixable.

find attached a warning.log file, for linux bitpit MPI shared compiling, gcc is 4.8.5, mpi  is 4.0.0(- gcc4.8.5 compliant)
[warning.log](https://github.com/optimad/bitpit/files/6090068/warning.log) 

Anyway, in any case I'm putting it in stand by, i need to investigate better this workaround for mpi compiling.

roccoarpa:
@andrea-iob 
After a deeper review, the proposed solution triggers strange behaviours in mpi compiling of an external project with bitpit. Have no clues how to control it. I think it's better to leave this specific workround be for now, if you agree. Sorry for the wasted time.

Anyway, about the proper warnings handling of warning.log file I can open a separate issue, if you want. Let me know. 

**Closing the current pull with wontfix/invalid**.

andrea-iob:
The warning you are seeing are only generated when using "old" compilers. Starting from gcc 4.9.0 those warnings are not issued anymore (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57211). That's why I never bother fixing them.

roccoarpa:
> The warnings you are seeing are only generated when using "old" compilers. Starting from gcc 4.9.0 those warnings are not issued anymore (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57211). That's why I never bother fixing them.

Ok, cool. Maybe leaving a notification on the bitpit Install.md should be enough, to avoid future misunderstandings.

PR #152, opened 2021-03-17
LA: allow to solve the transpose of a sparse system

andrea-iob:
Added a function to evaluate the transpose of a sparse matrix.


andrea-iob:
It is now possible to solve the transpose of a linear system.

marcocisternino (approved):

PR #158, opened 2021-04-12
patchkernel: fix evaluation of face vertex ids for polygons and polyhedra

andrea-iob:
There was a bug in the evaluation of local face vertex list for  for polygons and polyhedra.


roccoarpa (approved):
works on my tests. Fine for me.

PR #189, opened 2021-06-05
patchkernel: optimize update of adjacencies when multiple matches are allowed

andrea-iob:
Optimize update of adjacencies when multiple matches are allowed.


andrea-iob:
I've fixed an assert, now all tests pass.

roccoarpa (approved):

marcocisternino:
I tested this pull on engine test case. It works.

PR #190, opened 2021-06-09
levelset: fix evaluation of normal on border edges

andrea-iob:
Border edges doesn't have a neighbour.


marcocisternino (approved):

marcocisternino:
Tested further on engine case. Already tested on two-fluid case initialization. I confirm my approval.

PR #197, opened 2021-06-11
voloctree: optimize update of adjacencies

andrea-iob:
Moving the declaration of the two vectors outside the for loops we can avoid some allocations.

(The branch "1.7_voloctree.reduce.allocations" contains the same changes for bitpit v1.7.)


andrea-iob:
I've reworked the update of the adjacencies, the new algorithm should use less memory.

marcocisternino (approved):
Same mesh of the master and 1 minute less over ~20 in mesh generation

andrea-iob:
I've pushed the changes also on the 1.7 branch.

PR #199, opened 2021-06-21
IO module: add JSON support to configuration package

roccoarpa:
The following pull adds into IO::ConfigParser class the possibility to parse/write json formatted dictionaries. 
Data structure and interfaces of the Config/Config parser classes are the same as before, XML parsing/writing functionalities are also preserved and totally independent from the new json add-on.

RapidJSON external library is used. The external dep is handled silently during bitpit compiling, applying the following the scheme:
1) RapidJSON is found quietly on the system
2) if it's found, set to true a bitpit internal-only precomp variable HAS_RAPIDJSON_LIB and compile with json support.
3) otherwise set to false HAS_RAPIDJSON_LIB, and ignore JSON related code.
In any case, XML support is always guaranteed.
Dictionary format conversion is possible too (parse a JSON file and write it in XML and vice-versa).

In order to comply with the IO::Config tree structure, bitpit JSON parsing  is limited only to arrays of JSON objects. Arrays of doubles, integers, booleans, string or null are not supported, and any attempt to parse them trigger an error.  

A test_io_00007 is added to check json parsing and writing. 

Tested on gcc 4.8 Linux  and MinGW Win10 gcc-10

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -446,7 +446,7 @@ if (BITPIT_ENABLE_MPI)
 endif()
 set(OPERATORS_EXTERNAL_DEPS "")
 set(CONTAINERS_EXTERNAL_DEPS "")
-set(IO_EXTERNAL_DEPS "LibXml2")
+set(IO_EXTERNAL_DEPS "LibXml2;RapidJSON")
 set(COMMUNICATIONS_EXTERNAL_DEPS "MPI")
 set(LA_EXTERNAL_DEPS "PETSc")
 set(SA_EXTERNAL_DEPS "")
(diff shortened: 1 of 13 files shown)


andrea-iob (changes requested):
I've added some fixup commits, mainly changes to adapt the coding style to the one used in bitpit. I've also modified the functions that read/write JSON values, they are now more similar to each other and more similar to what is done in the XML configuration. Finally, I moved the read/write buffers into the heap. 

I would like to request two changes:
 - would it be possible to use a broader check for the required version (e.g., 1.1 or even 1)?
 - could you modify the .travis.yml file and add rapidJSON to the packages installed in the test environment?

roccoarpa:
To address requests: 
   1) There's practically a unique stable version 1.1.0 for now and almost all major linux distros support 1.1.0 in their repo packages. Older versions seems outdated.   
   2) for Ubuntu platforms there is a rapidjson-dev package available.  rapidjson v.1.1.0 is supported regularly from Ubuntu 18.04 Bionic up to 21.04. Current Xenial (16.04) in use on travis has the package, but it seems an outdated version. Maybe it will be better to insert the rapidjson testing as soon as the travis reference platform is upgraded.

andrea-iob (approved):

andrea-iob:
When no RapidJSON was found on the system, the test was exiting without calling MPI_finalize. This made the test fail on Travis. I've modified the test to call MPI finalize also when RapidJSON is not found.

PR #201, opened 2021-06-25
voloctree: avoid using cell interfaces during mesh update 

andrea-iob:
Mesh update was relying on cell interfaces to identify the vertices that should be deleted. This was unintended and now should be fixed.


marcocisternino (approved):

andrea-iob:
I've pushed the commits also on the 1.7 branch.

PR #204, opened 2021-07-07
patchkernel: optimize transfer of vertices/cells/interfaces into the patch

andrea-iob:
It is not possible to directly transfer a vertex/cell/interface into the storage. First a new element should be created and then that element can be replaced with the source that needs to be transferred into the patch. In order to improve performances, the created element can be just a dummy element, no need to copy its properties from the source.

### DIFF
--- a/src/patchkernel/cell.cpp
+++ b/src/patchkernel/cell.cpp
@@ -260,19 +260,6 @@ void Cell::_initialize(bool interior, bool initializeNeighbourhood, bool storeNe
 
 	// Neighbourhood
 	if (initializeNeighbourhood) {
-		// To reduce memory fragmentation, destroy both interfaces/adjacencies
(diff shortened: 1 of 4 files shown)


andrea-iob:
There was a problem with the updated of the bounding box, now it should be fixed.

edoardolombardi (approved):

roccoarpa (approved):

roccoarpa:
@andrea-iob: i've experienced some discrepancies when I tried to serialize a distributed mesh. Inconsistencies seems triggered by one of these commits. I'm trying to narrow the problem, I'll keep you posted.

roccoarpa:
UPDATE: last commit solve missing id trashing when deleting a ghost vertex. This solves the anomaly in serialization

roccoarpa (approved):

PR #205, opened 2021-07-08
travis: thin out test matrix

andrea-iob:
Reducing the number of cases will allow us to reduce the time spent running Travis jobs.


andrea-iob:
I'm merging this to avoid using too many Travis credits.

PR #208, opened 2021-07-15
levelset: support distributed objects

andrea-iob:
The current process may only have the portion of the object needed for evaluating the levelset on the cells of its mesh, therefore, when the bounding box of the object is needed, we need to evaluate the overall bounding box across all process.

@edoardolombardi I've also modified the levelset to handle object ids using an IndexGenerator. This should fix the problem you were having when deleting objects from the levelset.


edoardolombardi (approved):
Review focused on IndexGenerator introduction.

marcocisternino (approved):

PR #211, opened 2021-07-27
PatchKernel: fix  missing implementation of getInternalVertexCount method

roccoarpa:
method declared in hpp, but missing implementation in cpp.


andrea-iob:
I've moved the function where all other vertex-related function are defined.

andrea-iob (approved):

PR #212, opened 2021-07-29
discretization: modify how singular values are discarded

andrea-iob:
Threshold that defines when singular values are discarded is reduced and discarded singular values are now set to zero.

It seems that the code:
```
    for (int i = 0; i < k; ++i) {
        double sigma_plus = (m_sigma[i] > zeroThreshold) ? (1. / m_sigma[i]) : 0.;
        cblas_dscal(m, sigma_plus, &m_U[i*m], 1);
    }
```
corresponds to equation 24 of the paper and not to equation 22 (I haven't modified the comment, because I'm not sure if I'm right).

@marcocisternino A threshold of 1e-14 seems quite small, can this threshold be increased to 1e-12?

### DIFF
--- a/src/discretization/reconstruction.cpp
+++ b/src/discretization/reconstruction.cpp
@@ -2810,7 +2810,7 @@ void ReconstructionKernel::display(std::ostream &out, double tolerance) const
 /*!
  * Is the threshold below which a singular value is considered zero.
  */
-const double ReconstructionAssembler::SVD_ZERO_THRESHOLD = 1e-9;
+const double ReconstructionAssembler::SVD_ZERO_THRESHOLD = 1e-14;
 
 
 /*!
--- a/src/discretization/reconstruction.cpp
+++ b/src/discretization/reconstruction.cpp
@@ -3283,7 +3283,7 @@ void ReconstructionAssembler::computePseudoInverse(int m, int n, double zeroThre
     // To check if a signular value is non zero, we only need to check if it's
     // greater than the defined threshold.
     for (int i = 0; i < k; ++i) {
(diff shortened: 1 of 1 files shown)


marcocisternino:
Setting a threshold to an arbitrary value is more in the mood of numerically interpreting equation 22. To do what equation 24 prescribes, h value should be computed to have a meaningful and robust truncation, but it is not easy.

About the value of the zero threshold, I cannot argue the magnitude of tiny singular values depending on the geometrical configuration of the reconstruction. 1.0e-14 is one order of magnitude more than the machine epsilon and it can triggers ill-conditioned diagonal matrices.
If no communication is set to make the user aware of very tiny singular values, this threshold can consider more cases without destroying the reconstruction for sure. Conversely, having very tiny non-zero SV can be a problem if noise is in the right hand side, therefore we can hope to have clean RHS.

The possible solutions are to compute h or to expose the parameter to the user communicating the presence of discarded SV. The first is tricky, the second is unpractical. 

We can increase it and test cases where the reconstructions can give tiny singular values, to see if the threshold is good for the worst case scenario. 

I'm sorry about the duplicated PR, but I thought I was meant to push this modification having found the problem.
I'm going to delete my PR.

andrea-iob:
If we are using equation 22 the check should be abs(sigma_plus) > epsilon and not sigma_plus > epsilon (unless sigma_plus is always positive).

andrea-iob:
Singular values are non-negative real numbers. I'll add a comment.

andrea-iob:
Comment added.

marcocisternino (approved):

PR #214, opened 2021-07-30
levelset: fix identification of external cells

andrea-iob:
A cell should be considered external if it is completely outside the object bounding box. If the cell may be intersected by the object (i.e., the bounding box of the cell intersects the bounding box of the object), the cell cannot be flagged as external.


andrea-iob:
The centroid of the highlighted cell is inside the bounding box of the geometry, however the centroid of its face neighobur is outside. If the identification of the external cells is done using the centroid of the cell, the highlighted cell will be identified as internal and its face neighbours as external. The sign of the highlighted cell will then be defined by sign propagation, however that cell cannot be reached by sign propagation because it has not internal face neighbours.

levelset_problem

marcocisternino (approved):

PR #219, opened 2021-08-25
levelset: reduce memory usage

andrea-iob:
Levelset can now use two types of storages: sparse or dense. When sparse storage is used, objects will only allocate space for storing information of cells that are inside the narrow band. Instead, when dense storage is used, objects will still evaluate information only for cells inside the narrow band, but they will allocate space for all cells of the mesh. If the narrow band covers
a large portion of the domain, it will be more efficient to allocate space for all the cells of the mesh, rather than keep track of the cells inside the narrow band and store space only for those cells.

Selection of storage type is done during the construction of a levelset object and cannot be changed later.

On Cartesian patches, dense mode will store levelset data using arrays instead of PiercedStorages. This allows to evaluate the levelset also on Cartesian patches in memory-light mode (sign propagation is still not supported when a Cartesian patch is in memory-light mode, but now it should be easy to add an ad-hoc sign propagator for Cartesian patches).

This changes also reworks how levelset information are stored in order to reduce memory footprint of segmentation objects.

Changes are only lightly tested. I've added some tests for the new features, but some changes (e.g., dump/restore) are not yet been tested for regressions.


andrea-iob:
This week I would like to merge this pull request, ant objections?

PR #223, opened 2021-08-30
voloctree: simplify neighbour search

andrea-iob:
Use PABLO function for searching neighbours.

PABLO function for neighbour search has been simplified and micro optimized.


edoardolombardi (approved):

marcocisternino (approved):
Tested using CFD solver. Non-conformal mesh has been used.

PR #224, opened 2021-08-31
patchkernel: fix sorting of vertices

andrea-iob:
There were two problems with vertex ordering:
 - it is not possible to use a tolerance when sorting points, otherwise the final result will depend from the initial order. To identify the order of the two points, their coordinates will now be compared without any tolerance. However, before comparing the coordinates, it is necessary to check if the two points are coincident within the specified tolerance;
 - for polyhedral elements, the list of vertices should contain the vertices sorted in the same order they first appear in the face stream. This guarantees that vertices will be sorted in the same order regardless of the id of the vertices. In this way given an element, the list of its vertices generated by different processes will iterate the vertices in the same order.


roccoarpa (approved):
everything is ok for me.

roccoarpa:
@andrea-iob is it possible to quickly merge this pull?

roccoarpa:
Hello @andrea-iob, always me. I don't know if there is something else going on about this pull, but if not, can we merge it? Let me know.

andrea-iob:
Merged.

PR #235, opened 2021-10-28
patchkernel: add a constant max length in skd tree

edoardolombardi:
Initialize the skdBox extrema to values that do not trigger Floating Point overflow when squared.


roccoarpa:
mimic side: this pull close #https://github.com/optimad/mimic/issues/199

edoardolombardi:
Closing. Addressed in #240.

PR #237, opened 2021-10-31
patchkernel: fix evaluation of cell rank from consecutive id

andrea-iob:
This fixes the identification of cell rank from global/consecutive id.


andrea-iob:
Duplicate of #236.

PR #238, opened 2021-11-02
Reviewed guide README_WINDOWS.md

roccoarpa:
Latest version of MSYS2 (July 2021) provides binary packages for petsc 3.15.4 in both static and shared form (.a, dll.a/dll).
This simplifies a lot the  bitpit installation procedure on Windows under the MSYS2/MinGW64 environment.

The manual was updated introducing this new possibility. The old "building from sources" method was kept to provide guide to install older versions of petsc.

bitpit compliancy with petsc-3.15.4 was verified, in both MPI and Serial versions.


andrea-iob (approved):
I've replaced "(f.e. ..." with "(e.g., ..." (as far as I know the "for example" abbreviation is e.g.).

PR #239, opened 2021-11-03
cmake: fix definition of BITPIT_CMAKE_CONFIG_PATH_LAST variable in BITPITConfig.cmake.in

roccoarpa:
When bitpit is used as dependency library, BITPIT_CMAKE_CONFIG_PATH_LAST chached internal variable wasn't declared properly. This triggered the reconfigure procedure every time.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,4 +1,4 @@
-# BITPIT installation
+# bitpit installation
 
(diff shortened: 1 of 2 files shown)


andrea-iob (approved):
I've also fixed the case of bitpit name in the same file (it's bitpit, not BITPIT).

PR #250, opened 2021-11-25
patchkernel: implement move constructor/assignment

andrea-iob:
Implement move constructor and assignment operators.


andrea-iob:
@lucask81 ProxyVector move assignment operator was not properly generated on old versions of gcc due to a bug (see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60796 and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57728). I've added an explicit definition of that operator (and also for the move constructor).

lucask81 (approved):

marcocisternino (approved):

PR #251, opened 2021-12-03
PABLO: fix evaluation of tree constants

andrea-iob:
The constant MAX_LEVEL defines the maximum allowed level, hence tree constants need to be evaluated also for this level.


andrea-iob:
There is still an error that need to fixed. I will make an updated pull request in a couple of hours. Please wait before testing.

marcocisternino:
I would add a commit who definitely set the MAX_LEVEL to its real limit equal to 21.
A level more could not change too much, but it is more coherent with the use of a double precision Morton index

marcocisternino (approved):
I approve the fix, no matter what the decision about MAX_LEVEL value will be

andrea-iob:
I've fixed the size of the arrays that store the constants. Should be good to test now.

PR #252, opened 2021-12-05
PABLO: optimize refinement

andrea-iob:
New octants are created directly inside the container, whereas previously were created in a temporary storage and then moved
into the container.


edoardolombardi (approved):

marcocisternino (approved):
Tested with CFD solver on non-conformal mesh.

PR #253, opened 2021-12-05
PABLO: increase maximum allowed level for 2D trees

andrea-iob:
Two-dimensional trees will now use a two-dimensional Morton number, this allows to increase their maximum allowed level to 31.

The maximum allowed level is evaluated taking into account the following constrains:
 - the Morton number should be able to index all the possible octants of the tree, for a n-dimensional tree, the size, expressed in bytes, of the Morton number should be at least n time the maximum level;
 - the length of the tree is stored in an unsigned long integer, this means that the maximum level should be less that 32;
 - coordinates are stored using unsigned long integers (32-bit), we want to combined them in an unsigned long long integer (64-bit) to generate a unique key that identifies the vertices inside the tree. In order for this to be possible for all the coordinates of a n-dimensional tree, the maximum number of usable bits in a coordinate is 64 divided by n, this means that the maximum level should be less 64 divided by n.


andrea-iob:
We had a case were 20 levels were not enough (in that specific case, there was no need to use all the 31 levels, I think the refinement needed about 23 levels).

If I'm not missing anything, once the maximum level is evaluated as a function of the dimension, customizing the evaluation of the Morton number for the different dimensions should be just a matter of passing the tree dimension to the function that does the evaluation (and implementing the needed customized functions, but those function were already there). The additional maintenance overhead introduce by these changes seems acceptable to me.

Here the stats:
```
 src/PABLO/LocalTree.cpp      |  48 ++++++++++++++--------------
 src/PABLO/Map.cpp            |  12 ++++---
 src/PABLO/Octant.cpp         | 166 ++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------------------
 src/PABLO/ParaTree.cpp       |  43 +++++++++++++------------
 src/PABLO/morton.hpp         | 189 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-------------------
 src/PABLO/tree_constants.cpp |  31 +++++++++---------
 src/PABLO/tree_constants.hpp |  13 ++++----
 7 files changed, 315 insertions(+), 187 deletions(-)
```

There are 128 new added lines and some of them are comments.

After this changes, the maximum level is not a compile time constant anymore, but with some refactoring I think I can declare it constexpr (if this may be useful). However, even if it's not a compile time constant, the maximum level is still stored in the same data structure (TreeConstants), along with all other tree "constants" (none of which are compile time constants).

marcocisternino:
Some details about Morton number.
The procedure to calculate the Morton number is called inter-leaving, because binary representations of integer coordinates are inter-leaved to get the binary Morton number that we convert to a decimal 64bits number, as a quasi-unique (element tree level is needed) persistent label for a mesh element.
The largest integer on current machines is uint64_t, which is an unsigned integer, therefore all the bits are for the number representation, all the 64bits, no sign is needed.
In 3D 3 coordinates identify 1 element, therefore because of inter-leaving 64/3 = 21 (integer division) and we can use up to 21 bits per coordinate. In 2D for the same reasons we can use up to 32 bits per coordinate. About the 32bits of the coordinate c++ variable (uint32_t), in 2D all the 32 bits can be used, in 3D only the first 21 the other 11 are never used and are always zeors.
Finally the number of bits per coordinate defines the maximum integer distance between the root and the leaves of the tree in terms of tree levels.
Considering the PR, I agree with @edoardolombardi, I cannot see so important consequences in increasing the max level of the quadtree up to 32, but the case triggering this pull. However, as @andrea-iob remarked maintenance overhead is not huge. Finally, I have no special problem in changing max level in 2D, let's just do it as it should, 32 in 2D and 21 in 3D. We could plan the extension of PABLO towards a forest of PABLOs in a close future, orchestrated by the unstructured grid, or the Cartesian one to start. It is a lot of work, this is sure.

andrea-iob:
The maximum Morton number is not an hard-coded constant, it's evaluated taking into account the constraints it should fulfill:
```
/**
* Compute the maximum allowed level.
*
* The maximum allowed level is evaluated taking into account the following
* constrains:
*  - the Morton number should be able to index all the possible octants
*    of the tree, for a n-dimensional tree, the size, expressed in bytes,
*    of the Morton number should be at least n time the maximum level;
*  - the length of the tree is stored in an unsigned long integer, this
*    means that the maximum level should be less that 32;
*  - coordinates are stored using unsigned long integers (32-bit), we want
*    to combined them in an unsigned long long integer (64-bit) to generate
*    a unique key that identifies the vertices inside the tree. In order for
*    this to be possible for all the coordinates of a n-dimensional tree, the
*    maximum number of usable bits in a coordinate is 64 divided by n, this
*    means that the maximum level should be less 64 divided by n.
*
* \param dimension is the dimension of the space
* \result The maximum allowed level.
*/
inline int8_t computeMaximumLevel(uint8_t dimension)
{
    if (dimension == 0) {
        return 0;
    }

    // Evaluate the maximum level taking into account the size of the Morton
    int8_t level = (8 * sizeof(uint64_t)) / dimension;

    // Limit required for storing the length of the tree
    level = std::min(static_cast<int8_t>((8 * sizeof(uint32_t)) - 1), level);

    // Limit required for generating the XYZ key of the coordinates
    level = std::min(static_cast<int8_t>((8 * sizeof(uint64_t)) / dimension - 1), level);

    return level;
}
```

marcocisternino (approved):

andrea-iob:
I updated the documentation of the function "computeMaximumLevel" as requested.

PR #254, opened 2021-12-06
discretization: low order cell basis can be evaluated on all type of cells

andrea-iob:
Polynomials of degree 0 and 1 are supported on all type of cells, for pixels and voxels also degree 3 is supported.

Note: coefficients are evaluated as the volume average of the Taylor series expansion coefficients. In order to support higher order degrees on all type of cells, we will need to implement numerical integration of the expansions terms over the volume of the cell. This can be done using Gauss quadrature rules and evaluating the terms on the integration points using the function that evaluate the basis function on a specified point.

### DIFF
--- a/src/discretization/reconstruction.cpp
+++ b/src/discretization/reconstruction.cpp
@@ -307,7 +307,26 @@ void ReconstructionPolynomial::evalPointBasisDerivatives(uint8_t degree, uint8_t
 /*!
  * Evaluates the values of the basis for cell reconstruction.
  *
- * The method works only for ElementType::Voxel and ElementType::Pixel.
+ * Basis functions are evaluated as the volume average of the Taylor series
+ * expansion coefficients:
+ *
+ *              /
+ *      (1 / V) | (csi_0 * U_0 + csi_1 * U_1 + csi_2 * U_2 + ... ) dV
+ *              /
+ *
+ * where:
+ *
+ *   - csi_0 = 1;
+ *   - csi_1 = x;
+ *   - csi_2 = y;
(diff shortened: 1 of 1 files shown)


lucask81 (approved):

lucask81 (approved):

andrea-iob:
For some reasons the review done by @lucask81 is not enough to allow me to merge the pull request. @marcocisternino Can you have a look at these changes?

andrea-iob:
(We require a review from a user that has write access and @lucask81 doesn't.)

src/discretization/reconstruction.cpp
```diff
@@ -307,7 +307,26 @@ void ReconstructionPolynomial::evalPointBasisDerivatives(uint8_t degree, uint8_t
...
+ *   - csi_1 = x;
+ *   - csi_2 = y;
+ *   - csi_3 = z;
+ *   - csi_4 = 0.5 * (x - x_0)^2;
+ *   - csi_5 = 0.5 * (y - y_0)^2;
+ *   - csi_6 = 0.5 * (z - z_0)^2;
+ *   - ...;
+ *
+ * Polynomials of degree 0 and 1 are supported on all type of cells, for pixels
+ * and voxels also degree 3 is supported.
```
marcocisternino: There's a discrepancy between doc and code. Max supported degree in code is 2, while doc says 3.

marcocisternino: I pushed a fix

marcocisternino (approved):

PR #257, opened 2021-12-08
common: improve logger

andrea-iob:
The improvements are:
 - there are now 5 levels of message severity (DEBUG, INFO, WARNING, ERROR, and CRITICAL);
 - each severity level has an associated function that allows to write messages with the specified severity (log::debug(), log::info(), ...);
 - the function ```log::cout``` now accepts an argument that specifies the default severity level of the messages;
 - a new function to disable logging up to a specified severity level has been introduced (```void Logger::disable(log::Level level = log::CRITICAL)```);
 - some functions have been renamed (e.g., ```setVerbosity``` renamed to ```setDefaultVerbosity```, ```setPriority``` renamed to ```setDefaultSevierity```), old functions are still there but they are marked as deprecated.

Here some examples:
```
log.cout().disableFile(); // Logging on file will be disabled
log::cout() << "Information message" << std::endl;
log::cout(log::CRITICAL) << "Critical message" << std::endl;
log::info() << "Information message" << std::endl;
log::debug() << "Debug message" << std::endl;

log.cout("logger1") << disableConsole(log::INFO); // Only messages with severity greater than INFO will be printed on the console
log.cout("logger1", log::ERROR) << "Error message" << std::endl;
log.warning("logger1") << "Warning message" << std::endl;

log.cout("logger2") << disable(); // Disable logging (verbosity is overridden in order to suppress all output)
log.error("logger2") << "Error message" << std::endl; // This message will not be logged
log.cout("logger2") << disable(log::NOTSET); // Re-enable the logging (log::NOTSET will remove the verbosity override)
log.error("logger2") << "Error message" << std::endl;

```


andrea-iob:
I moved insertion/extraction operators from the operators module to the common module, because those operations need to be defined before the logger. If the operators are defined after the logger, they will not be found by the compiler when selecting the right specialization to be used for the insertion operator in Logger::println (gcc will still accept the code, but this is a gcc bug, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51577).

roccoarpa (changes requested):
Hi, new logger interface is fine for me, there are some minor fixes to do, I'm just listing them here:

- [ ] **.kdev4** folder, **code.kdev4** file and **cloc** perl script, i think they are accidentally here 
- [ ] example: pod_application_example_00002  does not compile and it's still using deprecated logger methods 
- [ ] (just a warning for future rebasing) when this branch is rebased on the actual master, newest test_LA_parallel_00002 will trigger the deprecated stuff too.

andrea-iob:
All fixed, thanks. I've also rebased the branch on top of the current master.

marcocisternino:
Could it be useful for users to rely on an "enable" method for the logger? I understand the semantic, but double negation to enable (or re-enable) the logger (at its level( it seems not intuitive to me. Maybe a kind of "enableDefault", even if I do not like "Default".

roccoarpa:
Hello guys, modifications are fine for me, but honestly I've not completely understood the request from Marco, and i don't want to rush my final approval. I will let you to iterate and elaborate more the concept, first.

andrea-iob:
> Could it be useful for users to rely on an "enable" method for the logger? I understand the semantic, but double negation to enable (or re-enable) the logger (at its level( it seems not intuitive to me. Maybe a kind of "enableDefault", even if I do not like "Default".

The function ```disable(log:Level level)``` will  provide an overriding verbosity level which takes precedence over the logger’s own level. Its effect is to disable all logging calls of severity level and below. When no arguments are specified, the overriding verbosity will be set to CRITICAL, hence all logger output will be disabled.

Its main purpose is to temporary throttle logging output down. When the logger verbosity can be set again to its defualt value, the function ```debug``` is called with the argument ```log::NOTSET``` and the verbosity override is removed.

(I've updated the wording of the example in the pull request description to better reflect what the function ```disable``` does.)

marcocisternino (approved):

marcocisternino:
Fine. Doc in sources was clear and your final comment too. Thanks for re-wording. I approve.

roccoarpa (approved):
Fine for me too

edoardolombardi (approved):

PR #258, opened 2021-12-17
System solver constructor from PETSc matrix dump file

marcocisternino:


andrea-iob (changes requested):
I've rebased the branch on the current master and fixed the conflicts.

Can we read, along with the matrix, also RHS and solution? The function SystemSolver::dump will dump also those vectors, but at the moment there is no way to import them back. The function can skip RHS and/or solution import if the corresponding file path is empty (matrix file path should be mandatory).

marcocisternino:
Absolutely yes. The way of doing it is exactly the same.
We could:
- check for the matrix path to exist and throw if not
- check for the solution and rhs path to exist and call vectorsCreate if not or loading them if yes.
Do you agree?

andrea-iob:
If the user specifies the path of a file (both matrix and vector) we should throw an exception is the file is not accessible. For vectors only, we can accept empty paths, in that case vectors will be created from scratch and their dimensions will be compatible with the matrix read from file.

marcocisternino:
I've added the dump and restore of the system transpose flag. I removed the block size from restore method because it is useless and misleading. The commit before the last one shows why, you can checkout there and run integration_test_LA_parallel_00003. Once approved, last two commits can be merged together, keeping the message of the last one.

marcocisternino:
I reworked dump and restore methods to allow system component dump and restore separately.
The matrix dump can be very expensive and time wasting if only solution is needed.

andrea-iob:
Please put back the blockSize in the restore function. If you want a restore function with a block size equal to one add an overload without that argument. Without that argument it will not be possible to properly restore a system with a block size different than one.

andrea-iob:
> Please put back the blockSize in the restore function. If you want a restore function with a block size equal to one add an overload without that argument. Without that argument it will not be possible to properly restore a system with a block size different than one.

I'm assuming the function MatLoad can load dump files with a block size different that one. If that's not the case the change is fine, but I think the documentation should be updated.

andrea-iob:
Having the transposed flag restored from the dump will create problems: the system is created with a transpose flag and that can be changed without notice when the system is restored. There is not even a function that can use to check if the transpose flag is on.

src/LA/system_solvers_large.cpp
```diff
@@ -2137,30 +2178,89 @@ void SystemSolver::restore(const std::string &directory, const std::string &pref
...
+#endif
+
+    // Create the matrix with dummy block size
+    matrixCreate(1);
+
+    // Restore matrix
+    restoreMatrix(directory, prefix);
+
+    // Restore info
+    restoreInfo(directory, prefix);
```
andrea-iob, line 2254: Restore the information before anything else.

andrea-iob (approved):

marcocisternino:
I squashed requested changes in 2 commits overall and rebased on master.

marcocisternino:
I have a main using system restore on arbitrary previously dumped systems. I would not make a test from it but an example that can be used to solve dumped systems. If you agree I can work on the example to make it usable.

andrea-iob:
> I have a main using system restore on arbitrary previously dumped systems. I would not make a test from it but an example that can be used to solve dumped systems. If you agree I can work on the example to make it usable.

Sure, let's make it a new example.

andrea-iob:
The new commit should fix the serial tests.

andrea-iob (approved):
I've done some minor changes to the examples.

andrea-iob:
Is there any more work that needs to be done, or is this pull request ready to be merged?

PR #259, opened 2022-01-17
patchkernel: fix Doxygen documentation

andrea-iob:
Improve the documentation of Element::setId and Element::setPID methods.

### DIFF
--- a/src/patchkernel/element.cpp
+++ b/src/patchkernel/element.cpp
@@ -488,9 +488,12 @@ void Element::_initialize(long id, ElementType type, std::unique_ptr<long[]> &&c
 }
 
 /*!
-	Sets the ID of the element.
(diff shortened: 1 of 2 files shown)


kgkara (approved):

src/patchkernel/element.cpp
```diff
@@ -488,19 +488,24 @@ void Element::_initialize(long id, ElementType type, std::unique_ptr<long[]> &&c
 }
 
 /*!
-	Sets the ID of the element.
+	Gets the id that identifies the element.
```
DMurataj01, line 491: Should this comment say 'Set' ?

andrea-iob: Indeed. I will fix this before doing the merge. Thanks.

PR #260, opened 2022-01-19
patchkernel: fix dump/restore of index generators

andrea-iob:
Index generators were not properly restored when auto indexing was set to false.


marcocisternino (approved):

roccoarpa (approved):
Sorry for this late review. Everything is fine for me too.

PR #263, opened 2022-01-26
levelset: fix restore of sign storage

andrea-iob:
When restoring LevelSetSignedObjectInterface, the sign storage needs to be explicitly initialized.


andrea-iob:
I fixed one problem in the levelset update that was triggering an assertion when a parallel restart was performed. Since the fixed problem was affecting the mesh update before the restart, you may need to re-save the dump files in order for the parallel restart to work properly.

marcocisternino (approved):

PR #264, opened 2022-02-07
Minor fixes for non-standard features (MSVC preliminary)

roccoarpa:
This pull contains some minor fixes, mainly focused to the preliminary introduction of MSVC as official compiler of bitpit. 
Main contents are: 

- Avoid using Variable Array Length (VLA) in static allocation of c-arrays. This feature is non-standard for all c-compilers. The basic rule (shared by all c compilers) is to allocate c-array with a const array-size. If the array-size is not known, switch to dynamic allocation and party's over. VLA let GNU and LLVM tolerate this, MSVC is inflexible and return all compile errors.
- Use a unique interface to call lapack stuffs, that is lapacke, since it's the official c-interface to lapack and widely portable. 
- Some datatype casting (size_t to long) are safe into linux context, not into windows context (long is a 4-byte regular int under windows.)
- other miscellaneous fixes


src/voloctree/voloctree.cpp
```diff
@@ -1955,7 +1955,7 @@ void VolOctree::_updateAdjacencies()
 
 	// Face information
 	int nCellFaces = m_cellTypeInfo->nFaces;
-	uint8_t oppositeFace[nCellFaces];
+	uint8_t oppositeFace[6];
```
andrea-iob, line 1958: Can we do this ```const uint8_t *oppositeFace = m_tree->getOppface()``` instead? In this way we don't even need to copy the array.

src/RBF/rbf.cpp
```diff
@@ -644,7 +644,7 @@ int RBFKernel::solve()
     int lda     = nS;
     int ldb     = nS;
     int info;
-    int ipiv[nS];
+    int *ipiv = (int*) malloc( nS * sizeof(int));
```
andrea-iob, line 647: If possible, I would use an std::vector (also in the POD module). Using malloc may lead to leaks. For example, after this change, ipiv is leaked if LAPACKE_dgesv fails (at line 678 there is a return before freeing ipiv).

src/patchkernel/element.cpp
```diff
@@ -1224,7 +1224,7 @@ ConstProxyVector<long> Element::getVertexIds(ElementType type, const long *conne
 					continue;
 				}
 
-				std::size_t vertexSortIndex = uniqueVertexIds.size();
+				long vertexSortIndex = uniqueVertexIds.size();
```
andrea-iob, line 1227: I would fix the declaration of uniqueVertexIds, it should become ```std::unordered_map<long, std::size_t> uniqueVertexIds;``` (after that change, there is a long that should become a std::size_t at line 1242).

andrea-iob (changes requested):
If needed, I can do the suggested changes.

src/voloctree/voloctree.cpp
roccoarpa: Sure, on it.

src/RBF/rbf.cpp
roccoarpa: Sure, on it

roccoarpa: Handled

roccoarpa:
I've done the requested changes. Everything compiles fine and tests work. Asking again the review.

src/patchkernel/element.cpp
roccoarpa: Sorry @marcocisternino , already fixed following the Andrea's suggestion. The fix should be coherent now. Please refer to the newest code

marcocisternino (approved):

andrea-iob (approved):

andrea-iob:
Merged with a couple of comsetic changes.

PR #284, opened 2022-02-19
Add lapacke-based solveLU

marcocisternino:
Two commits are in this branch.
The first one introduces two new solveLU functions (one for solution only, the other one for solution and factorization). The number of rhs is limited to 1. Two extern constant have been introduced to set the matrix layout using lapack values and to avoid application to include lapacke headers.
The second commit fix the determinant evaluation by fixing the complement build and by calling the complement function with opportune arguments.

### DIFF
--- a/src/LA/system_solvers_small.cpp
+++ b/src/LA/system_solvers_small.cpp
@@ -43,9 +43,18 @@
 # include "multiplication.hpp"
 # include "system_solvers_small.hpp"
 
+# include "bitpit_private_lapacke.hpp"
+
 namespace bitpit{
 
 namespace linearalgebra{
+
+namespace constants{
+
+const int ROW_MAJOR = LAPACK_ROW_MAJOR;
+const int COL_MAJOR = LAPACK_COL_MAJOR;
+
+}
 
 /*!
  * @ingroup system_solver_small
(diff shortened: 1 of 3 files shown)


src/LA/system_solvers_small.cpp
```diff
@@ -353,6 +363,62 @@ backwardSubstitution(U, z, x);
...
+    This routine works on the entire matrix, the number of considered equations
+    is equal to the leading dimension of the matrix, the same is for the r.h.s.
+    Multiple r.h.s. are not allowed.
+    This function is intended for both solution and factorization computation.
+
+    \param[in] layout matrix layout (possible values are ROW_MAJOR and COL_MAJOR).
+    \param[in] matrixOrder number of equations in linear system.
+    \param[in,out] A in input the coeffs of square matrix (linear), in output the the factors L and U, A = P * L * U.
+    \param[in,out] B in input r.h.s. of the linear system, in output the solution.
+    \param[out] ipiv the pivot indices that define the permutation matrix P, see [1].
```
andrea-iob, line 405: I don't see where reference [1] is defined.

src/LA/system_solvers_small.cpp
```diff
@@ -353,6 +363,62 @@ backwardSubstitution(U, z, x);
...
+/*!
+    Solve a linear system using Lapacke_dgesv (partial pivoting
+    and LU factorization). Containers MUST be correctly sized.
+    This routine works on the entire matrix, the number of considered equations
+    is equal to the leading dimension of the matrix, the same is for the r.h.s.
+    Multiple r.h.s. are not allowed. Permutation matrix not provided.
+    This function is intended for solution computation ONLY.
+
+    \param[in] layout matrix layout (possible values are ROW_MAJOR and COL_MAJOR)
+    \param[in,out] A in input the coeffs of square matrix (linear), in output the permuted factors L and U.
```
andrea-iob, line 376: Can we use the same desciption for both the solveLU functions?

andrea-iob (approved):

src/LA/system_solvers_small.cpp
marcocisternino: Almost. See new push, please

src/LA/system_solvers_small.cpp
marcocisternino: I forgot to correct, referring Lapack Doc.

PR #285, opened 2022-03-08
Fix undefined reports from clang memory sanitiser

DMurataj01:


src/patchkernel/cell.cpp
```diff
@@ -634,13 +634,16 @@ long * Cell::getInterfaces(int face)
...
 {
-	long *faceInterfaces = getInterfaces(face);
-	int nFaceInterfaces = getInterfaceCount(face);
-	for (int i = 0; i < nFaceInterfaces; i++) {
-		if (faceInterfaces[i] == interface) {
-			return i;
-		}
-	}
+    int nFaceInterfaces = getInterfaceCount(face);
+    if (nFaceInterfaces > 0) {
```
andrea-iob, line 638: What clang doesn't know is that in all cases where faceInterfaces is null, nFaceInterfaces is equal to zero. Would it be possible to silence the warning issued by clang using an assert? Something like "assert(nFaceInterfaces == 0 || faceInterfaces);"

andrea-iob (changes requested):

src/patchkernel/cell.cpp
DMurataj01: https://github.com/optimad/bitpit/pull/304/files

DMurataj01: ^^ PR above fixes the root cause behind these undefined behaviours

DMurataj01:
PR not required anymore - root cause behind these issues - https://github.com/optimad/bitpit/pull/304.

src/PABLO/tree_constants.cpp
```diff
@@ -56,8 +56,8 @@ TreeConstants::instance(uint8_t dim) {
 const TreeConstants::Instances &
 TreeConstants::instances() {
 	static TreeConstants::Instances instances = {{
-		TreeConstants(0),
-		TreeConstants(0),
+		TreeConstants(2),
+		TreeConstants(2),
```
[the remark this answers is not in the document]
DMurataj01: @andrea-iob these issues were picked up from running the clang memory sanitiser. It may be worth integrating a memory sanitiser in your pipeline to pick these undefined behaviours up so we can remove them?

PR #293, opened 2022-04-24
levelset: add support for evaluating the levelset on unstructured grids

andrea-iob:
It is now possible to evaluate the levelset on unstructured grids.


edoardolombardi:
I don't know the relationship with the pull request, but (at least) tests 1,2,3,4,6 of surfunstructured and tests 1,4 of sunrfunstructured parallel fail. 
On master branch all of them pass.

andrea-iob:
There were some failures with the leveset tests that are now fixed. I don't see problems with the surfunstructured tests.

edoardolombardi (approved):
Just ran bitpit tests and reviewed the code. No used in an external application.

andrea-iob:
Rebased on current master and added support for generic segmentation elements (before the change, only triangles were supported for three-dimensional segmentations).

andrea-iob:
The new levelset test was failing because generalized barycentric coordinates cannot be evaluated for degenerate polygons
(i.e. polygons with consecutive collinear edges). We may need to introduce some checks in the CG module to find out if a polygon is degenerate. For now, I slightly modified the geometry of the test case to avoid degenerate polygons.

PR #295, opened 2022-04-29
Implement automatic mesh partitioning using METIS

lucask81:
With this branch is now possible to partition unstructured mesh by using METIS.


src/patchkernel/patch_kernel_parallel.cpp
```diff
@@ -1750,15 +1657,231 @@ double PatchKernel::evalPartitioningUnbalance(const std::unordered_map<long, dou
...
+		std::vector<idx_t> nodeRanks(nNodes);
+
+		int status = METIS_PartMeshDual(&nElements, &nNodes, connectRanges.data(), connectStorage.data(),
+										elementWeights.data(), nullptr, &nCommonNodes, &nPartitions, nullptr,
+										nullptr, &objectiveValue, elementRanks.data(), nodeRanks.data());
+
+		if (status != METIS_OK) {
+			throw std::runtime_error("Error during partitioning (METIS error " + std::to_string(status) + ")");
+		}
+
```
roccoarpa, line 1743: Sorry for the dumb question,  but I don't know very well metis and just to foresee eventual bottlenecks in case I want to use directly this partitionPrepare Method with  cellweights. 
@edoardolombardi mimic used we used an external partitioner METIS_PartGraphKWay and we call a PatchKernel partition method employing a list of cellRanks to achieve distribution on ranks.  This bitpit version embeds a call to METIS_PartMeshDual instead, in the twin PatchKernel partition method using the cellWeights. There are substanstial differences between the 2? Advantages in using one instead of another?

roccoarpa (commented):
Aside the doubt(reported below), it's fine for me. I will stash mimic direct dependency from Metis when this pull will be merged.

src/patchkernel/patch_kernel_parallel.cpp
edoardolombardi: In mimic we use directly METIS_PartGraphKWay because we reconstruct the graph through the adjacencies structure of the patch. Inside METIS_PartMeshDual, I think it is done exactly the same (in a mesh-to-graph method, by supposing that two elements are neighbour when at least one edge is shared). 
In my opinion, the call to METIS_PartMeshDual can be useful if the adjacencies structure of the patch can be empty (@andrea-iob is it the case? What about the ghost layer/s creation during partition method?).
If here the adjacencies are always needed, the METIS_PartGraphKWay can be called directly and an internal double computation af tha adjacencies can be avoided.

andrea-iob: With METIS_PartMeshDual it's easier to handle cell weights.

PR #299, opened 2022-05-13
patchkernel: set alteration flags only when needed

andrea-iob:
If no adjacencies/interfaces are build, there is no need to set the associated alteration flag. Since alteration flags are stored in an unordered_map, this will improve PatchKernel memory usage when no adjacencies/interfaces are needed.


marcocisternino (approved):
Tested with dual solver porous tanks. It preserves the results of the master.

PR #300, opened 2022-05-16
patchkernel: make updateOwner method of PatchKernel public

edoardolombardi:
The method is useful to update the ownership of a parellel patch outside a complete update of the patch.

For example when filling a sub-selection of a _parallel_ mesh the selected ghost cells have to be connected to the local cells on each partition, for this a deletion of non-connected ghost cell must be performed; in order to carry out this identification/deletion procedure the adjacencies must be computed; when your starting mesh is completely owned by a single processor (_parallel_ but not _distributed_) the ghosts cleaning procedure is no needed and this allows to avoid the allocation of the adjacencies (lower memory usage, shorter execution time). 

The call to the method _updateOwner_ updates the (unitialized) owner information and a subsequent call to _isDistributed_ function gives the correct information. 

Using them during a patch filling method avoids to reimplement the same code in an external application.

### DIFF
--- a/src/patchkernel/patch_kernel.hpp
+++ b/src/patchkernel/patch_kernel.hpp
@@ -686,6 +686,7 @@ public:
 
 	bool isDistributed() const;
 	int getOwner() const;
+    void updateOwner();
 
 	void setHaloSize(std::size_t haloSize);
 	std::size_t getHaloSize() const;
@@ -1013,8 +1014,6 @@ private:
 
 	void updateGhostVertexOwners();
 	void updateGhostVertexExchangeInfo();
-
-	void updateOwner();
 
 	std::unordered_map<long, int> evaluateExchangeVertexOwners() const;
 #endif


andrea-iob:
The update() method should already update the owner information. If you want to be able to call the update method on a non-distributed mesh and have the partitioning-related information updated without the need of building the adjacencies, we can update _findGhostCellExchangeSources to not require the adjacendies when the mesh is not distributed.

edoardolombardi:
We need to understand if the patch is distributed or not during a filling procedure to decide a-priori if we need to run a cleaning orphan ghosts procedure or not. In case that some ghosts are orphan (selections can provide this case), the update method of the patch doesn't work, reaching (as expected) a dead-lock during some communication algorithm.

In order to update entirely the patch (method update(), but I suppose that is valid also for the method _findGhostCellExchangeSources) the ghosts have to be correctly built; in order to have the ghosts correctly cleaned and built we have to run the indentification of unconnected ghosts. This call (and the adjacencies computing) can be avoided if the sub-selection patch is not distributed. isDistributed method of the patch gives a wrong result if the ownership is not updated (through updateOwner method). 

The methods to retrieve this information correctly via the patch are already implemented, we have just make one of them public without any other modifications or new implementations, I think. 
If it's a problem (I don't see what's the problem, but maybe I miss something), we have no issue to implement it externally, it was just to avoid the duplication of the same code.

edoardolombardi:
An alternative can be to insert in the isDistributed method a flag that the user can be set to true/false if he wants to ask for the collective information (updateOwner method in practice) or the stored one. In this way also a non-updated mesh can be queried.
This is the implementation that we have currently in mimic.

PR #304, opened 2022-05-17
containers: Fix undefined behaviour when flatVector is empty but has items

DMurataj01:


src/containers/flatVector2D.tpp
```diff
@@ -1091,6 +1091,10 @@ const T* FlatVector2D<T>::operator[](std::size_t i) const
 {
     assert(isIndexValid(i));
 
+    if(m_v.empty()) {
```
andrea-iob, line 1094: nit: could you add a space between if and the parenthesis to be consistent with the code-style of this file?

src/containers/flatVector2D.tpp
```diff
@@ -1106,6 +1110,10 @@ T* FlatVector2D<T>::operator[](std::size_t i)
...
```
andrea-iob, line 1113: nit:  space between if and the parenthesis.

andrea-iob (approved):
It would be great if you could add the prefix "containers: " to the commit message. This makes easier to spot commits related to a specific module.

I will add this fix also in the upcoming 1.8 version.

PR #306, opened 2022-05-22
patchkernel: reduce memory usage of Skd-Tree cache

andrea-iob:
Cell centroids were only used to detect element ordering. For this purpose, the centroid of the element's bounding box can be used.

In order to recover the performances lost due to the computation of the cell bounding box centroid, I've also optimized the generation of the Skd-Tree. The splitting between left and right elements can be done evaluating only the position of the element along the direction of the splitting rather than evaluating the whole bounding box of the element.


marcocisternino (approved):
Tested with dual solver porous tanks. It preserves the results of the master. Memory reduction is very light in this case with triangulations of bodies allocating a peak of ~180MB and of regions allocating a peak of ~1GB

PR #307, opened 2022-05-22
patchkernel: allow to pass a generic container to the functions that delete cells/vertices/interfaces

andrea-iob:
This avoid creating a list of vertices to delete during the adaption of the VolOctree patch.

### DIFF
--- a/src/patchkernel/patch_kernel.tpp
+++ b/src/patchkernel/patch_kernel.tpp
@@ -165,7 +165,7 @@ bool PatchKernel::deleteInterfaces(const IdStorage &ids)
 	// sure to delete the last interface just once (after deleting the last
 	// interface, another interface becomes the last one and that interface
 	// may be on the deletion list as well, and on and so forth).
-	std::size_t lastId;
+	long lastId;
 	if (!m_interfaces.empty()) {
 		lastId = m_interfaces.back().getId();
 	} else {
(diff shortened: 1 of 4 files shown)


src/patchkernel/patch_kernel.tpp
```diff
@@ -45,6 +45,150 @@ std::unique_ptr<patch_t> PatchKernel::clone(const patch_t *original)
...
+		return false;
+	}
+
+	// Deleting the last interface requires some additional work. If the ids
+	// of interfaces to be deleted contain the last interface, that interface
+	// is deleted after deleting all other interfaces. In this way we make
+	// sure to delete the last interface just once (after deleting the last
+	// interface, another interface becomes the last one and that interface
+	// may be on the deletion list as well, and on and so forth).
+	std::size_t lastId;
```
DMurataj01, line 168: think this lastId should be of type `long` / `id_t`

andrea-iob: Will fix it, thanks.

marcocisternino (commented):
Tested with dual solver porous tanks. It preserves the results of the master.
I could approve but I saw a comment asking for modification. I will approve when the modification is introduced.

src/patchkernel/patch_kernel.tpp
andrea-iob: Fixed.

marcocisternino (approved):

Issue #269, opened 2022-02-11
IO/configuration: XML/JSON update dependencies


roccoarpa:
Consider to use the more portable boost xml/json interfaces (rapidxml/rapidjson) to get rid of the actual libxml2/rapidjson dependencies. 
In particular, libxml2 is uber-supported on Linux systems, but quite rare on Windows native ones. 

Understand if it's more convenient to switch to the newest and drop the older ones, or to keep both for a whhle for a safe redundancy. 
issue connected to #268 activity

PR #309, opened 2022-05-24
IO/config: parse from XML/JSON formatted string

roccoarpa:
Added Config specialization class (ConfigStringParser) to read and write XML (or JSON) directly from and onto c++ string. The class is useful to exhange info without passing from filesystem each time.

Enhanced IO integration tests 2 e 4 to prove the capabilities of the new class.

### DIFF
--- a/src/IO/configuration.cpp
+++ b/src/IO/configuration.cpp
@@ -374,50 +374,46 @@ namespace config {
 /*!
     Construct a new parser.
 
-    \param XMLorJSON false to activate XML format, true for JSON format if supported,
-    otherwise fallback automatically to XML.
+    \param format type of format to parse, of ConfigStringParser::Format enum type
 */
-ConfigStringParser::ConfigStringParser(bool XMLorJSON)
+ConfigStringParser::ConfigStringParser(ConfigStringParser::Format format)
     : Config(false)
 {
 #if HAS_RAPIDJSON_LIB
-    m_xmlOrJson = XMLorJSON;
+    m_format = format;
 #else
-    BITPIT_UNUSED(XMLorJSON);
+    BITPIT_UNUSED(format);
     //revert to default XML
-    m_xmlOrJson = false;
+    m_format = Format::XML;
 #endif
 }
 
 /*!
     Construct a new parser, activating multiSections support.
 
-    \param XMLorJSON false to activate XML format, true for JSON format if supported,
-    otherwise fallback automatically to XML.
+    \param format type of format to parse, of ConfigStringParser::Format enum type
     \param multiSections if set to true the configuration parser will allow
     multiple sections with the same name
 */
-ConfigStringParser::ConfigStringParser(bool XMLorJSON, bool multiSections)
+ConfigStringParser::ConfigStringParser(ConfigStringParser::Format format, bool multiSections)
     : Config(multiSections)
 {
 #if HAS_RAPIDJSON_LIB
-    m_xmlOrJson = XMLorJSON;
+    m_format = format;
 #else
-    BITPIT_UNUSED(XMLorJSON);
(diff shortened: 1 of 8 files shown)


src/IO/configuration.cpp
```diff
@@ -360,4 +361,113 @@ namespace config {
...
+*/
+
+
+/*!
+    Construct a new parser.
+
+    \param XMLorJSON false to activate XML format, true for JSON format if supported,
+    otherwise fallback automatically to XML.
+*/
+ConfigStringParser::ConfigStringParser(bool XMLorJSON)
```
andrea-iob, line 380: Since this is a new class and there are no compatibility concerns, can we use an enum rather than a bool? It will make the code easy to understand (```ConfigPartser(true)``` vs ```ConfigParser(ConfigParser::FORMAT_JSON)```)

src/IO/configuration_XML.cpp
```diff
@@ -283,6 +319,75 @@ void writeConfiguration(const std::string &filename, const std::string &rootname
...
 
+/*!
+    Write the Config Tree to a c++ string (xml-stringfication). All contents will 
+    be appended to the target source string.
+    The method is meant for general-purpose xml info flushing.  
+    \param source string to write to
+    \param rootConfig pointer to the Config tree to be stringfied.
+    \param rootname (optional) name of the root section. Default is "root".
+*/
+void writeBufferConfiguration(std::string &source, const Config *rootConfig, const std::string &rootname)
```
andrea-iob, line 330: Would it be possible to add a private function that takes in input a xmlTextWriterPtr? This will allow to share code between writeConfiguration and writeBufferConfiguration? For "symmetry", this apply also to "readBufferConfiguration".

In bitpit , arguments that are modified by the function are usually passed by pointer at the end of the argument list.

Do we really need a function with a different name? The signature of writeConfiguration and writeBufferConfiguration is different, we can use the same name.

I would still pass a "format" argument (also in the read* function). In this way the function that write the configuration to file and the one that write the configuration to string will generate the same output.

src/IO/configuration_JSON.cpp
```diff
@@ -246,6 +282,43 @@ void writeConfiguration(const std::string &filename, const Config *rootConfig, b
...
+
+    NOTE: JSON root document object does not have a key name unlike XML.
+
+    Tree contents will be written to a plain c++ string, with UTF-8 standard encoding, and 
+    appended to any previous content of the source string.
+    
+    \param[in] source string to write JSON contents.
+    \param[in] rootConfig pointer to the Config tree to be written on target string
+*/
+void writeBufferConfiguration(std::string &source, const Config *rootConfig)
```
andrea-iob, line 296: Arguments that are modified by the function are usually passed by pointer at the end of the argument list.

src/IO/configuration.cpp
```diff
@@ -360,4 +361,113 @@ namespace config {
 
 }
 
+/*!
+    \class ConfigStringParser
```
andrea-iob, line 365: Is this new class really needed? Can we just add an overload to the read/write functions of ConfigParser, this overloads will take in input the string and the format?

src/IO/configuration_XML.cpp
roccoarpa: You're right a compact code is always better. But since pending pull https://github.com/optimad/bitpit/pull/267 and needs of issue #269 suggest a major underhood rework of tree interface and its parsing capabilities, and the introduction of rapidXML will make a lot easier to compact the code w.r.t. to libxml2, I think it's not the case to lose much time on this now. Anyway, as you prefer, if you want to fix it right now, i'll fix it. Let me know

roccoarpa:
moved elsewhere.

PR #310, opened 2022-05-30
patchkernel: allow to build more than one layer of ghost cells

andrea-iob:
It is now possible to build more than one layer of ghost cells.

Draft because it needs more testing.

### DIFF
--- a/src/patchkernel/patch_kernel.hpp
+++ b/src/patchkernel/patch_kernel.hpp
@@ -499,17 +499,22 @@ public:
 	CellIterator addCell(ElementType type, std::unique_ptr<long[]> &&connectStorage, long id = Element::NULL_ID);
 #if BITPIT_ENABLE_MPI==1
 	CellIterator addCell(const Cell &source, int owner, long id = Element::NULL_ID);
+	CellIterator addCell(const Cell &source, int owner, int haloLayer, long id = Element::NULL_ID);
 	CellIterator addCell(Cell &&source, int owner, long id = Element::NULL_ID);
+	CellIterator addCell(Cell &&source, int owner, int haloLayer, long id = Element::NULL_ID);
 	CellIterator addCell(ElementType type, int owner, long id = Element::NULL_ID);
+	CellIterator addCell(ElementType type, int owner, int haloLayer, long id = Element::NULL_ID);
 	CellIterator addCell(ElementType type, const std::vector<long> &connectivity, int owner, long id = Element::NULL_ID);
+	CellIterator addCell(ElementType type, const std::vector<long> &connectivity, int owner, int haloLayer, long id = Element::NULL_ID);
 	CellIterator addCell(ElementType type, std::unique_ptr<long[]> &&connectStorage, int owner, long id = Element::NULL_ID);
+	CellIterator addCell(ElementType type, std::unique_ptr<long[]> &&connectStorage, int owner, int haloLayer, long id = Element::NULL_ID);
 #endif
 	bool deleteCell(long id);
 	template<typename IdStorage>
 	bool deleteCells(const IdStorage &ids);
 #if BITPIT_ENABLE_MPI==1
 	CellIterator ghostCell2InternalCell(long id);
-	CellIterator internalCell2GhostCell(long id, int owner);
+	CellIterator internalCell2GhostCell(long id, int owner, int haloLayer);
 #endif
 	virtual double evalCellSize(long id) const = 0;
 	BITPIT_DEPRECATED(long countFreeCells() const);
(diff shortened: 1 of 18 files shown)


lucask81 (approved):

andrea-iob:
@edoardolombardi @roccoarpa This pull request make mandatory specifying the ghost layer when adding new ghost cells. Old methods still exists, but they are marked as deprecated. However, I've just realized that the signature of the new methods may confuse the compiler (calls to the old method can replaced with calls to the new one). Can you check if this changes work fine in your applications?

edoardolombardi:
@andrea-iob just as update: we didn't check yet this modifications. 
Sorry, for the delay; I think that we have mainly to check if we can force the single layer of ghosts easily and in a clean way for the user. 
In this moment the whole mimic project works and is based on the assumption that a single layer of ghosts is allowed; some methods and algorithms are developed on this hypothesis.   

As soon as possible we'll try the branch and test mimic considering that the deprecated methods cannot longer be used.

src/patchkernel/patch_kernel_parallel.tpp
```diff
@@ -0,0 +1,85 @@
...
+ */
+template<typename ExcludeList>
+bool PatchKernel::confirmCellHaloLayer(const Cell &cell, int haloLayer, const ExcludeList &excludeList) const
+{
+	// Search if a face neighbour is in the previous halo layer
+	//
+	// We first search in the face neighbours only, because its faster than
+	// searching among all the neighbours.
+	//
+	// Searching for neighbours in the previous layer works also for ghosts in
```
edoardolombardi, line 48: What about a ghost that is in the first layer but it's an edge/vertex-only neighbor of inner cells (the same for computeCellHaloLayer method)?
If I'm not wrong in the first layer of ghosts also the ghost-cells that are neighbors of the locale ones only through a vertex or an edge (no faces) are included. 
Maybe, in order to be consistent also the edge and vertex neighbors should be considered during the propagation algorithms to compute the n-layers halo (i.e. processCellsNeighbours method should work not only for face neighbors)?

src/patchkernel/patch_kernel.hpp
```diff
@@ -497,17 +497,22 @@ friend class PatchManager;
 	CellIterator addCell(ElementType type, const std::vector<long> &connectivity, long id = Element::NULL_ID);
 	CellIterator addCell(ElementType type, std::unique_ptr<long[]> &&connectStorage, long id = Element::NULL_ID);
 #if BITPIT_ENABLE_MPI==1
-	CellIterator addCell(const Cell &source, int owner, long id = Element::NULL_ID);
-	CellIterator addCell(Cell &&source, int owner, long id = Element::NULL_ID);
-	CellIterator addCell(ElementType type, int owner, long id = Element::NULL_ID);
-	CellIterator addCell(ElementType type, const std::vector<long> &connectivity, int owner, long id = Element::NULL_ID);
-	CellIterator addCell(ElementType type, std::unique_ptr<long[]> &&connectStorage, int owner, long id = Element::NULL_ID);
+	BITPIT_DEPRECATED(CellIterator addCell(const Cell &source, int owner, long id = Element::NULL_ID));
```
edoardolombardi, line 500: Can we maintain all the deprecated methods with no haloLayer argument?  

They are useful overloading methods with default 0-value of haloLayer that I prefer to save for a couple of reasons:

- the library keeps its compatibility with the previous version and all the applications based on bitpit are saved (they are currently forcely using only the first layer of ghosts);
- these methods remains consistent with all the other methods with default value of halo size/layer, as for instance the constructor of SurfUnstructured/VolUnstructured where this default value exists.

edoardolombardi: Just a note, but @andrea-iob you know because your ongoing review, please consider to rebase on #340 and to use messages to indicate alternative methods.

Mostly when multiple functions are deprecated all together, for a bitpit user can be useful to have a message directly during the compilation with the function that he has to use instead of to have to navigate inside the headers.

src/patchkernel/patch_kernel_parallel.tpp
andrea-iob: The function confirmCellHaloLayer will first check among the face vertices, if a neighbour that belongs to the previous layer is not found there, it will extend the search also to node and edge neighoburs. The first check is much faster than the latter and will cover most of the cases, therefore I thought it was a good idea to perform the "extended" check only when needed.

edoardolombardi: Sorry, it was commented futher below in the code.

src/patchkernel/patch_kernel.hpp
andrea-iob: > Can we maintain all the deprecated methods with no haloLayer argument?

Done. However, the two set of function have very similar signatures, this might generate problems.

edoardolombardi (approved):

marcocisternino (approved):

marcocisternino:
I tested by writing an example building a volunstructured in parallel directly and checking the goodness of the source and target lists. I confirm that it is possible to build a patch providing rank sub-domains with ghosts in more than a halo layer. I tested 2 layers.

PR #312, opened 2022-05-31
patchkernel: fix evaluation of partitioning information

### LINKED ISSUE #311: Bitpit Debug MPI : assertion fails in PatchKernel::getOwner method
In MPI, debug mode, integration_test_volunstructured_00004 crashes with a failed assertion while calling PatchKernel::setPartition method. 
The problem is caused by assert at **row 211** of patchkernel_parallel.cpp (getOwner). 

Question: is this assert even useful? Logical if from row 212 seems robust enough to me. 
As it is, this assert is a booby trap, best to remove it.

andrea-iob:
The assert in Patchkernel::getOwner was wrong, additionally partitioning information status should be tracked also when there is only one processor.

Close #311

### DIFF
--- a/src/patchkernel/patch_kernel_parallel.cpp
+++ b/src/patchkernel/patch_kernel_parallel.cpp
@@ -3938,8 +3938,6 @@ bool PatchKernel::arePartitioningInfoDirty(bool global) const
 {
 	if (!isPartitioned()) {
 		return false;
-	} else if (getProcessorCount() == 1) {
-		return false;
 	}
 
 	bool partitioningInfoDirty = m_partitioningInfoDirty;
@@ -3959,8 +3957,6 @@ bool PatchKernel::arePartitioningInfoDirty(bool global) const
 void PatchKernel::setPartitioningInfoDirty(bool dirty)
 {
 	if (dirty && !isPartitioned()) {
-		return;
-	} else if (getProcessorCount() == 1) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


src/patchkernel/patch_kernel_parallel.cpp
```diff
@@ -208,12 +208,13 @@ bool PatchKernel::isDistributed(bool allowDirty) const
...
 {
-	assert(allowDirty || arePartitioningInfoDirty(false));
-	if (!allowDirty || !arePartitioningInfoDirty(true)) {
-		return m_owner;
-	} else {
+	if (allowDirty) {
 		return evalOwner();
 	}
+
+	assert(!arePartitioningInfoDirty(false));
```
roccoarpa, line 215: Another booby trap. If you're working  MPI with 1 processor, isPartitioned is always true (that means communicator is always set by class constructor), arePartitioningInfoDirty return always true. See below

src/patchkernel/patch_kernel_parallel.cpp
```diff
@@ -3937,8 +3938,6 @@ bool PatchKernel::arePartitioningInfoDirty(bool global) const
 {
 	if (!isPartitioned()) {
 		return false;
-	} else if (getProcessorCount() == 1) {
-		return false;
```
roccoarpa, line 3941: This was the failsafe for the assert failing before.

roccoarpa (changes requested):

roccoarpa (approved):

kgkara (approved):

PR #314, opened 2022-06-06
PABLO: Reduce some UBSAN `-fsanitize_implicit_conversion` runtime errors.

DMurataj01:


DMurataj01:
Note - the idim conversions are for consistency with dim also being a int8_t

andrea-iob (approved):
On new code we try to use C++ static_cast<> rather than C-style cast, but there is no need to updated this pull request. Thanks for the changes.

DMurataj01:
> On new code we try to use C++ static_cast<> rather than C-style cast, but there is no need to updated this pull request. Thanks for the changes.

Following the existing code style but noted for next time!

PR #317, opened 2022-06-25
voloctree: reduce visibitiy of PABLO messages

andrea-iob:
In release mode, VolOctree patch will display PABLO messages only when their severity is WARNING or higher.

To achieve this I had to rework a bit the logger: loggers that were sharing the same buffer were sharing also the visibility settings, but this was not the intended behaviour. Other than fixing this problem, the reworked logger should be faster.

### DIFF
--- a/src/PABLO/Octant.cpp
+++ b/src/PABLO/Octant.cpp
@@ -32,6 +32,8 @@
 #include <cmath>
 #include <iostream>
 
+namespace bitpit {
+
 /*!
  * Input stream operator for class Octant. Stream cell data from memory
  * input stream to container.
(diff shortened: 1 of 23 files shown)


andrea-iob:
I had to make some changes to make the code work with GCC 12 (and hopefully clang, tests will confirm this). Now, stream operators for bitpit classes are declared inside the bitpit namespace. Since at least one of the arguments passed to those stream operators are in the bitpit namespace, the compiler is still able to pick the correct overload thanks to ADL (Argument-dependent lookup);

marcocisternino:
Can you rebase on master, please? Thanks!

marcocisternino (approved):

PR #321, opened 2022-06-26
levelset: check if there are seeds to be used for sign propagation

andrea-iob:
We need at least one seed to be able to perform sign propagation (i.e., at least one cell should be in the levelset narrowband).

### DIFF
--- a/src/levelset/levelSetSignPropagator.cpp
+++ b/src/levelset/levelSetSignPropagator.cpp
@@ -204,6 +204,21 @@ void LevelSetSignPropagator::propagate(const LevelSetObjectInterface *object, Le
         }
     }
 
+    // Check if there are seeds to be used for sign propagation
+    //
+    // We need at least one seed to be able to perform sign propagation (i.e., at least one
+    // cell should be in the levelset narrowband).
+    long nGlobalSeeds = rawSeeds.size();
+#if BITPIT_ENABLE_MPI
+    if (m_mesh->isPartitioned()) {
+        MPI_Allreduce(MPI_IN_PLACE, &nGlobalSeeds, 1, MPI_LONG, MPI_SUM, m_mesh->getCommunicator());
+    }
+#endif
+
+    if (nGlobalSeeds == 0) {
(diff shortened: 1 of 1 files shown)


src/levelset/levelSetSignPropagator.cpp
```diff
@@ -204,6 +204,21 @@ void LevelSetSignPropagator::propagate(const LevelSetObjectInterface *object, Le
...
+    // cell should be in the levelset narrowband).
+    long nGlobalSeeds = rawSeeds.size();
+#if BITPIT_ENABLE_MPI
+    if (m_mesh->isPartitioned()) {
+        MPI_Allreduce(MPI_IN_PLACE, &nGlobalSeeds, 1, MPI_LONG, MPI_SUM, m_mesh->getCommunicator());
+    }
+#endif
+
+    if (nGlobalSeeds == 0) {
+        throw std::runtime_error("Unable to propagate the sign: the list of seeds is empty!");
```
edoardolombardi, line 219: What about to force the computation of the levelset value on a single cell just to retrieve the sign?

I dont know if I well understood, but I'm thinking to a possible workflow where the levelset is used to perform boolean operations on several geometries. 
One of the geometries can be something like a design space, a volume defining the working domain; in order to guarantee robustness the computation should run also if this volume is bigger than the computing mesh; for instance an user, to be sure that the whole mesh is included in the design space, could define the computing mesh smaller than the design space.

In this case, as user I expect that the execution goes forward with a negative sign on the whole mesh for the levelset related to the design volume.

What do you think?

andrea-iob: My goal was just to make clear that, in the current implementation, we need a seed (cell in the narrowband) to propagate the sign. Without these changes, the error in the case no seeds were present was something like "Mismatch in sign propagation".

Updating the sign propagation and evaluate the sign of the external region with an explicit calculation of the distance between one external cell and the geometry is on my TODO list. It should be simple, however I need to check if the object that does the propagation is able to evaluate the distance (I think it isn't, because that object doesn't have access to the levelset object, we may need some preliminary changes).

edoardolombardi (approved):

PR #322, opened 2022-07-12
PABLO: ParaTree cleanup octant looping in adaption routines

DMurataj01:


src/PABLO/ParaTree.cpp
```diff
@@ -3698,9 +3698,8 @@ namespace bitpit {
     ParaTree::adaptGlobalRefine(bool mapper_flag) {
         //TODO recoding for adapting with abs(marker) > 1
         uint32_t nocts0 = getNumOctants();
-        vector<Octant>::iterator iter, iterend = m_octree.m_octants.end();
 
-        for (iter = m_octree.m_octants.begin(); iter != iterend; ++iter){
+        for (auto iter = m_octree.m_octants.begin(); iter < m_octree.m_octants.end(); ++iter){
```
andrea-iob, line 3701: This code has a couple of downside with respect to the original one:
 - it will re-evaluate end() at each iteration;
 - it's less generic, because not all iterators implement the "<" and ">" operators.

If you want to get rid of iterator declarations I would suggest using range-based for loops:
```
for (Octant &octant : m_octree.m_octants) {
```
or, if range-based for loops are not an option, declare a variable for the end iterator in the for construct and use the "!=" operator:
```
for (auto iter = m_octree.m_octants.begin(), iterend = m_octree.m_octants.end(); iter != iterend; ++iter){
```

DMurataj01: Nice catch

PR #323, opened 2022-07-20
levelset: having no seeds is fine if no sign propagation is needed

andrea-iob:
Throwing an exception when there are no seed is wrong if sign propagation is not needed.


andrea-iob:
@lucask81 This may fix the problems you ave with bitpit.

lucask81 (approved):
@andrea-iob you are right this commit fix the problem with the immerflow test cases.

PR #324, opened 2022-07-29
Add a function to find duplicate cells

andrea-iob:
Identification of duplicate cells is based on vertex ids: a cell is a duplicate of another cell if it shares the same vertex ids.


edoardolombardi (approved):

marcocisternino (approved):
I tested using voloctree in CFD solver. No duplicated cells were expected. No issue observed.

PR #325, opened 2022-08-09
discretization: allow to define a pool for the weights

andrea-iob:
This will allow to re-use the weights and thus will reduce re-allocations.


marcocisternino (approved):

andrea-iob:
There is now a separate class for discrete stencils that want to use a memory pool. In this way, if memory pool is not needed, we don't need to store an extra pointer.

lucask81 (approved):
I approve since I tested and saw a reduction in the allocations.

PR #326, opened 2022-08-09
patchkernel: add alteration flags to the data structures that will be squeezed

andrea-iob:
This ensures that memory hold by empty alteration flags will be released when the patch is squzeed.


marcocisternino:
I tested it but I'm not a reviewer. Complex mesh generation test passes.

edoardolombardi (approved):

PR #327, opened 2022-08-10
PABLO: silence some "uninitialized variable" warnings reported by Coverity

andrea-iob:
They all seem false positive, with the exception for the initialization of the global number of octants.


marcocisternino (approved):

marcocisternino:
I just checked the code. It seems good to me.

PR #328, opened 2022-08-10
patchkernel: consolidate initialization of the communicator

andrea-iob:
There is now a function to initialize a dummy communicator when MPI support is disabled.


marcocisternino (approved):
I checked the code. No tests needed but unit ones, in my opinion.

PR #331, opened 2022-08-18
Optimize header inclusion

andrea-iob:
I've removed some unused headers and replaced others with forward declarations (this changes have been suggested by cppclean).


andrea-iob:
I fixed a problem with a missing header in the logger.hpp file.

marcocisternino (approved):

PR #305, opened 2022-05-17
patchkernel: flush id generators after renumbering

DMurataj01:

PR #332, opened 2022-08-18
patchkernel: reset index generators after renumbering cells/vertices/interfaces

andrea-iob:
The bug was found and fixed by @DMurataj01, this is just a slightly modified version of #305. 

@DMurataj01 Can you please check if this pull request fixes the problem?


DMurataj01:
Yep nice one - this also fixes the bug 👍

andrea-iob:
I'm merging this.

PR #333, opened 2022-08-21
levelset: change synchronization method of narrowband cache 

andrea-iob:
The algorithm used for updating the narrow band adds the cell to the cache and immediately after it evaluates cache information. In this case using a concurrent synchronization for the PiercedStorages is faster than using a journaled synchronization.


marcocisternino (approved):
My complex geometry meshes generation tests pass. I cannot see evident saving in time, but results are correct.

PR #335, opened 2022-08-28
PABLO: simplify 2:1 balance

andrea-iob:
I've rewritten some functions to reduce code duplication. The refactoring removed 1300 lines of code.

There are also some performance improvements. The biggest one is related to the last commit. After the latest code refactoring, it is no more necessary to relay the AUX bit for 2:1 balance. Moreover, that bit was not properly reset and 2:1 balance was check on many more octants than actually needed. This leads to a speed improvement of 25x for the function that perform load balance. All tests pass, however a thorough code review and some more test are needed to make sure the changes are fine.

Original profile of localBalance:
original

Updated profile of localBalance:
updated


marcocisternino (approved):
I tested using several complex geometries. Basically, the generated mesh is equal to the mesh generated by the previous version. The max speed up I could see in mesh generation is about 5% of the previous version. The tests have been based on: number of global elements, checking 1:1 of slices of the mesh, comparison by subtraction of global ids, checking of number of the elements and of their configuration level by level.
We must continuously pay the our attention to the effects of this mod, because a lot of unknown configurations should be tested.

andrea-iob:
Rebased on current master.

PR #336, opened 2022-09-08
patchkernel parallel: update isPartitioned() to return false for 1 MPI rank

DMurataj01:


andrea-iob:
Considering the patch partitioned also when running on a single MPI process was done on purpose. This allows to test MPI code in the very simple case of just one process. I am open to change this, however can you share some details on the segmentation fault you are seeing when running on a single process?

DMurataj01:
> Considering the patch partitioned also when running on a single MPI process was done on purpose. This allows to test MPI code in the very simple case of just one process. I am open to change this, however can you share some details on the segmentation fault you are seeing when running on a single process?

Would love to help you but this particular issue was seen by a colleague. I think it makes sense to be able to set n procs to 1 when benchmarking and it follows a serial routine. - @andrea-iob maybe have a debugging override ?

andrea-iob:
This change has been merged through #365.

PR #338, opened 2022-09-12
common: fix timestamp displayed by the logger

andrea-iob:
The string where the timesamp was build was too small.


marcocisternino (approved):
Tested in several situations

PR #339, opened 2022-09-14
 discretization: fix logic to detect if polynomial storage should be resized

andrea-iob:
The current storage size was not properly evaluated.

I've also updated the function that evaluates the polynomial coefficients: since the weight matrix is usually small, it's faster perform matrix vector multiplication using an ad-hoc code rather than using CBLAS functions..

### DIFF
--- a/src/discretization/reconstruction.cpp
+++ b/src/discretization/reconstruction.cpp
@@ -530,6 +530,7 @@ void ReconstructionPolynomial::clear(bool release)
     m_nFields = 0;
 
     if (release) {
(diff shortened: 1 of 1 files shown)


andrea-iob:
I've put back the CBLAS function in the evaluation of the polynomials. The raeson I was seeing poor performances for small matrices may be related to the fact that I was using an unoptimized BLAS library.

andrea-iob:
Rebased on current master.

marcocisternino (approved):

PR #340, opened 2022-09-19
common: add message to deprecated macros

edoardolombardi:
Added new macro to define functions as deprecated with message as input argument.

With this new macros BITPIT_DEPRECATED_FOR (one for each supported compiler as the original ones) is possibile to pass a message to indicate wich method has to be used instead.

Currently compiled and tested only for GNUC case.

### DIFF
--- a/src/common/compiler.hpp
+++ b/src/common/compiler.hpp
@@ -80,8 +80,8 @@ do {                  \
 #   define BITPIT_DEPRECATED(func) func __attribute__ ((deprecated))
 #   define BITPIT_DEPRECATED_FOR(func, replacement) func __attribute__ ((deprecated(" Use " #replacement)))
 #elif defined(_MSC_VER)
-    #define BITPIT_DEPRECATED(func) __declspec(deprecated) func
-    #define BITPIT_DEPRECATED_FOR(func, replacement), replacement __declspec(deprecated(" Use " #replacement)) func
+#   define BITPIT_DEPRECATED(func) __declspec(deprecated) func
+#   define BITPIT_DEPRECATED_FOR(func, replacement), replacement __declspec(deprecated(" Use " #replacement)) func
 #else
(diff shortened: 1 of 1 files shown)


src/common/compiler.hpp
```diff
@@ -75,10 +75,13 @@ do {                  \
  */
 #if defined __GNUC__
 #   define BITPIT_DEPRECATED(func) func __attribute__ ((deprecated))
+#   define BITPIT_DEPRECATED_FOR(func, replacement) func __attribute__ ((deprecated(" Use " #replacement)))
 #elif defined __clang__
 #   define BITPIT_DEPRECATED(func) func __attribute__ ((deprecated))
+#   define BITPIT_DEPRECATED(func, replacement) func __attribute__ ((deprecated(" Use " #replacement)))
```
andrea-iob, line 81: Maybe this should be "BITPIT_DEPRECATED_FOR"?

src/common/compiler.hpp
```diff
@@ -75,10 +75,13 @@ do {                  \
...
 #elif defined(_MSC_VER)
     #define BITPIT_DEPRECATED(func) __declspec(deprecated) func
+    #define BITPIT_DEPRECATED(func), replacement __declspec(deprecated(" Use " #replacement)) func
```
andrea-iob, line 84: Same here.

edoardolombardi: Sure, thanks. I put two bugs in three lines of code... 🗡️

andrea-iob:
Could you please also add "define BITPIT_DEPRECATED_FOR(func, replacement) func" in the else branch? This will allow unsupported compilers to still compile bitpit even if we don't explicitly define a deprecated macro for them.

andrea-iob (approved):

andrea-iob:
Since we are changing this part of the code, I changed the message displayed when deprecated macros are not implemented ad fixed the indentation.

PR #341, opened 2022-09-23
IO: miscellaneous fixes for the XML reader/writer

andrea-iob:
These are the same changes of #267 but without the part the add attributes support.


roccoarpa (changes requested):
compiling fails: 
```
 In member function ‘void bitpit::Config::dump(std::ostream&, int) const’:
.../bitpit/src/IO/configuration_config.cpp:405:41: error: ‘entry’ was not declared in this scope
  405 |             out << indent << padding << entry.first << " = " << entry.second << std::endl;
```
Please re-ask me the review when everything is fixed.

roccoarpa (approved):

PR #343, opened 2022-09-23
cmake: fix compilation when clang-format is not available

andrea-iob:
We need to check if clang-format executable has been found before using it.


marcocisternino (approved):
Tested in an environment without clang-format

PR #344, opened 2022-09-23
containers: simplify FlatVector2D::pushBack function

andrea-iob:
Let's leave to the standard library the burden of optimizing the insertion of new values.


marcocisternino (approved):
Code checked. No test more than integration tests is needed in my opinion.

PR #345, opened 2022-09-26
volunstructured: enable generation of multiple halo layers 

lucask81:
This PR changes the way the halo size is set. 
Now it is possible to set the halo size only if the patch is all on a single process. 
The function _getMaxHaloSize is overridden also for volunstructured, in this way it is possible to set more the one layer of ghost also for this patch.


andrea-iob:
I moved the changes that allows to set the halo layer on already partitioned patches in #310. This pull request is now only about enabling multiple halo layers on unstructured patches.

andrea-iob (approved):
Will be merged after #310.

PR #349, opened 2022-10-13
cmake: add a compiler flag to warn about missing override keywords

andrea-iob:
Warn about overriding virtual functions that are not marked with the override keyword.


marcocisternino (approved):
Tested with application using the same flag.

PR #359, opened 2022-10-26
CG: fix in the triangulation of a polygon

### LINKED ISSUE #355: Problem with bitpit::CGElem::distancePointPolygon  when compiling bitpit in Debug Mode
The problem regards the usage of bitpit::CGElem::distancePointPolygon  when compiling bitpit in Debug mode. An assertion is triggered by line /home/kostas/Optimad/bitpit/src/CG/CG_elem.cpp:66 because of line 572 of the same file. Here we need to project a point on the polygon in pink contour of this picture
image
when computing the minimum distance between the point and the surface.
On the right hand side you might see the vertex coordinates comprising this polygon.
The problem is that the formed triangle of the 3 pink colored points of this figure
image
has a zero surface area. You might see their coordinates on the right hand side.
However, when it comes to computation of the minimum distance of a point from the surface, this normally should not make the code to stop, because the neighbour boundary interfaces don't lead to triangles of zero surface area. So, they could provide the  minimum distance instead and the code could just neglect this problematic triangle. Alternatively, the triangles could be created with the barycenter of the polygon.

kgkara:
So far, this took place by creating triangles with the 1st vertex of the polygon and two other vertices of the polygon. In case these 3 vertices are collinear, the debug mode does not complete its execution. Now, the approximate barycenter of the polygon is computed and is used along with two vertices of the polygon to create the triangles.
Closes #355

### DIFF
--- a/src/CG/CG_elem.cpp
+++ b/src/CG/CG_elem.cpp
@@ -431,8 +431,15 @@ bool _intersectBoxPolygon(array3D const &A0, array3D const &A1, std::size_t nVS,
 
             int intrCount = partialIntr.size();
             for( int i=0; i<intrCount; ++i){
-
+                // Get the intersection point
+                //
+                // Polygon triangulation adds a dummy vertex at the centroid of the polygon (V0).
+                // This vertex should not be taken into account for intersection identification.
                 array3D &candidateCoord = partialIntr[i];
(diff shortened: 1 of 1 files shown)


andrea-iob:
I've updated the documentation, removed a warning and removed the usage of a couple of BITPIT_UNUSED on arguments that are now actually used.

I will merge after the tests finish.

andrea-iob (approved):

PR #360, opened 2022-11-05
LA: optimize update of system matrix

andrea-iob:
If the sizes of PETSc data types match the sizes of data types used by bitpit a direct update can be performed, otherwise the matrix is updated using intermediate data storages.


lucask81 (approved):
I tested the changes by using immerflow. In this case, the changes have a small contribution to the simulation performance. I approve this PR since all bitpit tests passed.

PR #363, opened 2022-11-11
discretization: add some fast paths for constant polynomials

andrea-iob:
It is possible to optimize the evaluation of constant polynomials taking advantage of the fact that basis value is 1 and basis derivative is 0. 

This kind of optimization was already done for the evaluation of the polynomial value and gradient. Let's extend it also for the update of the polynomials and for the evaluation of the weights.


lucask81 (approved):
I tested by running the immerflow test and the unstructured turbulent flat plate in order to check the results.

PR #364, opened 2022-11-13
LA: add a fast path for updating matrix values

andrea-iob:
The fast path will use the PETSc function MatSetValuesRow to update all the values of a row at once (without the need to get the row pattern).
    
Fast update can be performed if:
  - the matrix has already been assembled;
  - the assembler is providing all the values of the row;
  - values provided by the assembler are sorted by ascending column.


lucask81 (approved):
I tested these changes with Immerflow. The changes reduce the overall simulation time.

PR #367, opened 2022-11-30
LA: allow to set the block size of the linear systems

andrea-iob:
Rather than working with individual elements in the matrix, it is possible to employ blocks of elements. The size of the blocks can be defined during assembly. When a size different that one is provided, the matrix will store elements by fixed-sized dense nb × nb blocks, where nb is the size of the blocks. Blocking may be advantageous when solving PDE-based simulations that leads to matrices with a naturally blocked structure (with a block size equal to the number of degrees of freedom per cell).


andrea-iob:
Using block matrices can increase performances of PETSc matrix-vector product and PETSc triangular solves obtained from an ILU(0) factorization more than 50% (see https://petsc.org/main/developers/matrices/).

PR #369, opened 2022-12-21
voloctree: fix partitioning initialization

andrea-iob:
The container passed to the function that performs the initial load balance was not the container with the correct weights.

### DIFF
--- a/src/voloctree/voloctree_parallel.cpp
+++ b/src/voloctree/voloctree_parallel.cpp
@@ -45,12 +45,19 @@ void VolOctree::initializeTreePartitioning()
 	// Assigning a null weight to every octant but the last octant and then
 	// doing a load balance will have the effect of moving all the octants
 	// to the first process.
+	std::vector<double> octantWeightStorage;
+	const std::vector<double> *octantWeights;
+
 	std::size_t nOctants = m_tree->getNumOctants();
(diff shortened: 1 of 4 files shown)


src/voloctree/voloctree_parallel.cpp
```diff
@@ -45,12 +45,19 @@ void VolOctree::initializeTreePartitioning()
...
+	std::vector<double> octantWeightStorage;
+	const std::vector<double> *octantWeights;
+
 	std::size_t nOctants = m_tree->getNumOctants();
-	std::vector<double> octantWeights(nOctants, 0.);
-	if (nOctants > 0) {
-		octantWeights[nOctants - 1] = 1.;
+	if (nOctants > 1) {
+		octantWeightStorage.assign(nOctants, 0.);
+        octantWeightStorage[nOctants - 1] = 1.;
```
marcocisternino, line 54: indentation

marcocisternino (approved):
Code checked. Apparently a loadBalance without weights was performed always because of m_partitioningOctantWeights nullity

src/voloctree/voloctree_parallel.cpp
andrea-iob: Fixed.

PR #370, opened 2023-02-03
patchkernel: improve adaption/partitioning tracking

andrea-iob:
Adaption/partitioning tracking is extended also to vertices and interfaces. However, tracking of vertices and interfaces offers only a limited set of information.

During adaption, information available for tracking purposes are the following:
 - internal cells that have been coarsened/refined;
 - new internal vertices that have been created;
 - internal vertices that have been deleted;
 - new ghost cells that have been created;
 - new ghost vertices that have been created;
 - ghost cells that have been deleted;
 - ghost vertices that have been deleted;
 - new interfaces that have been created;
 - interfaces that have been deleted.

During partitioning, information available on the sender side for tracking purposes are the following:
 - internal cells that have been sent;
 - internal vertices that have been sent;
 - new ghost cells that have been created (some of the internal cells that
   have been sent may have become ghosts cells);
 - new ghost vertices that have been created (some of the internal vertices
   that have been sent may have become ghosts vertices);
 - ghost cells that have been deleted;
 - ghost vertices that have been deleted;
 - new interfaces that have been created;
 - interfaces that have been deleted.

During partitioning, information available on the receiver side for tracking purposes are the following:
 - internal cells that have been received;
 - internal vertices that have been received;
 - new ghost cells that have been created;
 - new ghost vertices that have been created;
 - ghost cells that have been deleted (some ghost cells may have been replaced by internal cells that have just been received);
 - ghost vertices that have been deleted (some ghost vertices may have been replaced by internal vertices that have just been received);
 - new interfaces that have been created;
 - interfaces that have been deleted.

PR #371, opened 2023-02-03
patchkernel: add the concept of "adaption mode"

andrea-iob:
Adaption mode tells if the patch can be adapted and which strategies can be used to adapt the patch.

The following adaption modes are supported:
 - disabled, no adaption can be performed;
 - automatic, adaption is performed specifying which cells should be refined/coarsen and then the patch will perform all the alterations needed to fulfill the adaption requests;
 - manual, this modes allows to use low level function to add and delete individual cells and vertices. It's up to the user to guarantee the consistency of the patch.

Adaption mode replaces the concept of "expert mode". The functions "isExpert" and "setExpert" are now deprecated.

Depends on #370.


src/lineunstructured/lineunstructured.cpp
```diff
@@ -74,15 +74,15 @@ LineUnstructured::LineUnstructured()
     among the processes
 */
 LineUnstructured::LineUnstructured(int dimension, MPI_Comm communicator)
-    : LineKernel(PatchManager::AUTOMATIC_ID, dimension, communicator, 1, true)
+    : LineKernel(PatchManager::AUTOMATIC_ID, dimension, communicator, 1, ADAPTION_MANUAL, PARTITIONING_DISABLED)
```
edoardolombardi, line 77: Why PARTITIONING_DISABLED?

andrea-iob: Fixed.

andrea-iob:
I've removed all the changes not directly related to this pull request.

marcocisternino (approved):
check code and tested in application using voloctree.
Tests using manual adapted volunstructured or surfunstructured would test the branch better than I did.

edoardolombardi (approved):

PR #375, opened 2023-03-01
patchkernel: add function to rotate the patch

andrea-iob:
The function allows to rotate a patch along a specified axis.


andrea-iob:
I removed a spurious unused argument in "Vertex::rotate".

edoardolombardi (approved):
We should check the validity of the inputs, for instance the axis points (if points n1 and n2 are equal the norm computation throws an exception, isn't it)?

PR #376, opened 2023-03-08
levelset: fix evaluation of levelset for surfaces that contain pixel elements

andrea-iob:
Connectivity of pixel elements does not list the vertices in anti-clockwise order. This needs to be properly taken into account when flipping the normal and when evaluating segment information.


andrea-iob:
I've added an additional fix for the hash operator of ElementHalfItem.

edoardolombardi (approved):

PR #377, opened 2023-03-13
levelset: derive segmentation objects from the corresponding kernel

andrea-iob:
Instead of storing the segmentation kernel in a member, let's just derive from it.

Depends on #376.


marcocisternino (approved):

edoardolombardi (approved):

andrea-iob:
Rebased on current master.

PR #378, opened 2023-03-20
patchkernel: fix evaluation of normal for polygonal elements

marcocisternino:
The return vector must be normalized to assure to be a versor.


andrea-iob:
I've added some comments.

andrea-iob (approved):

andrea-iob:
The branch "ai.element.eval.normal" contains a simplified version.

PR #381, opened 2023-04-01
Add missing Voloctree Mapper rank-overlapped ids getter

marcocisternino:
This branch:
- aligns rank-overlapped reference ids getter (getSentReferenceIds) to the mapped ids ones , by providing non-empty rank lists only
- changes the return datatype of these getters into an unordered_map
- introduces a missing getter, specifically the getReceivedReferenceIds method

It has been tested using rank/ids list getters for mesh mapping MPI communications in an application code.


edoardolombardi (approved):
Approved only by code review and run of bitpit tests (I've no application using that feature).

andrea-iob:
To be consistent with current git history, the portion of the commit title should only contain the name of the module (e.g., "voloctree: ...").

PR #382, opened 2023-04-03
patchkernel: fix evaluation of voxel volume

andrea-iob:
The formula for evaluating the volume needs the position of four vertices that identifies the x, y, and z edges.


edoardolombardi (approved):

andrea-iob:
Pull request #383 will introduce a test for the unstructured patch that checks the evaluation of cell volumes (admittedly, it will be a very basic test, but still an improvement over the current situation where there are no tests).

PR #383, opened 2023-04-03
 patchkernel: rework evaluation of characteristic size of three-dimensional elements

andrea-iob:
The updated formulas are more appropriate when the characteristic size is used for the evaluation of the timestamp in transient simulations. Also, the evaluation of the characteristic size for polygons and polyhedra was missing. 

When no better formula is known, characteristic size is evaluated as the volume divided by the area of the largest face (or by the area divided by the length of the longest face for two-dimensional elements).


andrea-iob:
In some contexts, the updated equations can still provide a size that is larger/smaller that desired. We may want to implement other formulas for those specific cases (for example, the levelset provides formulas for evaluating the radius of the tangent sphere or the radius of the bounding sphere).

edoardolombardi (approved):

marcocisternino (approved):

PR #384, opened 2023-04-04
voloctree: simplify adaption info emplace_back

DMurataj01:


src/voloctree/voloctree.cpp
```diff
@@ -1027,10 +1027,9 @@ std::vector<adaption::Info> VolOctree::_adaptionPrepare(bool trackAdaption)
 				adaptionInfo = &(adaptionData[adaptionInfoId]);
 			}
 
-			assert(adaptionInfo);
-			adaptionInfo->previous.emplace_back();
-			long &cellId = adaptionInfo->previous.back();
-			cellId = getOctantId(octantInfo);
+			if (adaptionInfo) {
+				adaptionInfo->previous.emplace_back(getOctantId(octantInfo));
+			}
```
andrea-iob, line 1032: The simplification is fine, however this portion of code requires a valid adaptionInfo pointer. If the pointer is not valid there is a bug somewhere and we cannot just ignore it.

Can you change the code as follows:
```
assert(adaptionInfo);
adaptionInfo->previous.emplace_back(getOctantId(octantInfo));
```

Also, the commit message is just ".", using something along the lines of the title of the pull request would be fine (e.g. "voloctree: simplify usage of adaption info emplace_back").

andrea-iob (changes requested):
I like the simplification, we just need to keep the assert.

andrea-iob (approved):

PR #386, opened 2023-04-04
voloctree: fix performance-move-const-arg clang tidy error in importCells (+ correct local typo)

DMurataj01:


andrea-iob:
Out of curiosity, do you have more information on why using a copy is better than using move?

DMurataj01:
> Out of curiosity, do you have more information on why using a copy is better than using move?

Not exactly sure, we have clang-tidy as part of our pipeline and this is something which I've had to fixup

PR #387, opened 2023-04-04
voloctree parallel: use std::make_unique

DMurataj01:


andrea-iob:
Unfortunately bitpit is limited to c++11 and make_unique is c++14.

PR #389, opened 2023-04-05
patchkernel: fix VTK parallel output

marcocisternino:
In 3D meshes, if at least one element without info (POLYGONS, POLYHEDRA) is present even in just one mesh partitions, all the ranks must write VTK DataArrays offset, types, connectivity, faces and faceoffsets in their own vtu file. Otherwise, single vtu files are readable but the pvtu file is not.


andrea-iob (approved):
I've added some comments.

PR #393, opened 2023-05-09
LA: fixes for serial builds

marcocisternino:
System Solver 's assembly methods have been fixed to build without MPI


andrea-iob:
In parallel builds, we should always explicitly pass a communicator during the assembly. Can you check if the branch "ai.LA.fix.serial.builds" works as well? That branch defines the functions that do not take the communicator in input only   for serial builds.

andrea-iob (approved):

roccoarpa (approved):
Seems ok to me.

PR #395, opened 2023-05-19
reconstruction: correct the computation of derivatives

ksamouchos:
Function "computeDerivatives" evaluates the gradient along the given normal direction similarly to function "computeGradients". However, when dimensions are zero, computeDerivatives throws a segmentation error, while computeGradients returns a zero value. This pull-request modifies computeDerivatives to be in accordance with computeGradients.

### DIFF
--- a/src/discretization/reconstruction.cpp
+++ b/src/discretization/reconstruction.cpp
@@ -800,6 +800,11 @@ void ReconstructionPolynomial::computeValues(int degree, const std::array<double
         return;
     }
 
+    // Early return if the polynomial is not initialized
+    if (m_dimensions == 0) {
+        return;
+    }
+
     // Get values
     const std::size_t fieldValuesStride = 1;
     double *fieldValue = values;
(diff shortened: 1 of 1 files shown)


andrea-iob:
Can you remind me what m_dimensions is?

andrea-iob:
In bitpit the commits should have the format "module_name: commit message".

ksamouchos:
> 

Done!

ksamouchos:
> In bitpit the commits should have the format "module_name: commit message".

Done!

andrea-iob (changes requested):
This only fixes the case when degree equal to one. The same function will fail some lines below if the degree is greater than one.

I've added a check on the number of dimensions in all functions that evaluates derivatives and values.

ksamouchos:
I incorporated changes recommended by @andrea-iob.

andrea-iob (approved):

PR #396, opened 2023-05-23
cicd: add github configuration to run continuous integration

andrea-iob:
Continuous integration is now handled by github actions and runs on non-cloud machines (Ubuntu 20.4).

CI works in the same way as before, there are just some differences in the tests being run (we don't test anymore the compilation using older gcc versions).

The results of the tests are available at the follwing link: "https://github.com/optimad/bitpit/actions/workflows/continuous_integration.yml". Here an example of how the test results look like:
bitpit_ci_1
bitpit_ci_2

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -1,0 +1,119 @@
+name: Continuous integration
+
+on:
+  workflow_dispatch:
+    inputs:
+  pull_request:
+    branches: [ "master" ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  setup:
+    runs-on: ubuntu
+    name: Setup
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build environment
+        shell: bash
+        run: |
+          DOCKER_BUILDKIT=1 docker buildx build \
+          --builder=container \
+          --cache-from type=local,src=/home/bob/products/bitpit/ubuntu-cache \
+          --cache-to type=local,dest=/home/bob/products/bitpit/ubuntu-cache \
+          --rm \
+          --target bitpit-environment \
+          -f environments/ubuntu/Dockerfile .
+      - name: Remove temporary docker images and containers
+        shell: bash
+        run: |
+          for CONTAINER in $(docker ps -a -q)
+          do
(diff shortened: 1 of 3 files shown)


andrea-iob:
I forgot to say that Travis will still be used for submitting Coverity analysis. It is possible to use github actions also for that task, but I will do the porting sometime in the (hopefully near) future.

katemay:
One thing just came to mind. Since this is open source, would it make sense to run this pipeline on github cloud instead of on bob? I read that github advises not using self-hosted runners for public repos to protect against pull requests from forks running potentially damaging code. Beyond that, isn't it free to use github cloud on an open source project? If bitpit is in the cloud, that's more resources for the other projects.

edoardolombardi:
> One thing just came to mind. Since this is open source, would it make sense to run this pipeline on github cloud instead of on bob? I read that github advises not using self-hosted runners for public repos to protect against pull requests from forks running potentially damaging code. Beyond that, isn't it free to use github cloud on an open source project? If bitpit is in the cloud, that's more resources for the other projects.

I completely agree

andrea-iob:
[Here](https://github.com/orgs/community/discussions/26722) there are some suggestions on how to safely enable self-hosted runners for public repositories. This are the two proposed solutions:
 - "The best (and most cumbersome) way to prevent this is to check Require approval for all outside collaborators in the repo settings as explained [here](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks). Then you have to manually approve workflows running on each PR, after you have first checked the PR for any changes to workflow files."
 - "Another way to mitigate the risk would be to have a private “companion” repo to your public repo that the self-hosted runner is assigned to as explained [here](https://www.ideasawakened.com/post/radauthenticator-part-4-build-automation-with-delphi-and-github-by-installing-a-self-hosted-runner) but I haven’t tried this myself."

The first solution seems the fine for us.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,86 @@
+name: Continuous integration
+
+on:
+  workflow_dispatch:
```
katemay, line 4: Would the workflow_dispatch be available to anyone on the internet since this is open source? It would probably be a bad idea to allow anyone in the world to run a job on bob.

andrea-iob: We can remove it, however I think that running a github action requires some kind of permission on the repository.

PR #400, opened 2023-05-31
levelset: fix evaluation of levelset on segmentation with pixel elements

andrea-iob:
There were still some cases where the pixel's vertices were not properly ordered before passing them to the CG utilities.

Depends on #394.

### DIFF
--- a/src/patchkernel/element_reference.cpp
+++ b/src/patchkernel/element_reference.cpp
@@ -1176,6 +1176,30 @@ double Reference2DElementInfo::evalPointDistance(const std::array<double, 3> &po
 }
 
 /*!
+    Check if the vertices are ordered counter-clockwise.
+
+    \result Return true if the vertices are ordered counter-clockwise,
+    false otherwise.
+*/
+bool Reference2DElementInfo::areVerticesCCWOrdered() const
+{
+    return true;
+}
+
+/*!
+    Get the index of the vertex occupying the n-th position in the
+    counter-clockwise ordered list of vertices.
+
+    \param n is the requested position
+    \result The index of the vertex occupying the n-th position in the
+    counter-clockwise ordered list of vertices.
+*/
+int Reference2DElementInfo::getCCWOrderedVertex(int n) const
+{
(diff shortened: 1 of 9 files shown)


src/patchkernel/volume_kernel.cpp
```diff
@@ -209,4 +209,99 @@ bool VolumeKernel::isPointInside(long id, double x, double y, double z) const
 	return isPointInside(id, {{x, y, z}});
 }
 
+/*!
+	Get the anti-clockwise ordered list of vertex for the specified face.
+
+	\param cell is the cell
+	\param face is the face
+	\result The anti-clockwise ordered list of vertex for the specified face.
```
marcocisternino, line 217: I would homogenize the use of anti-clockwise and counter-clockwise (or CCW)

marcocisternino:
We are using "anti-clockwise ordered", "counter-clockwise ordered" and "ordered" in doc, variable names and method prototypes.  I suppose all the three expressions mean exactly the same thing, i.e. "counter-clockwise ordered". If this is the case, couldn't we use the same expression everywhere? Even only "ordered" is good to me, being explained in high level doc that ordered vertices are always counter-clockwise ordered. I would preferer to be more clear by always specifying "counter-clockwise ordered" or "CCW" for sake of conciseness

src/patchkernel/element_reference.cpp
```diff
@@ -1511,6 +1539,36 @@ int ReferencePixelInfo::getCCWOrderedLocalVertex(int n) const
...
+bool ReferencePixelInfo::areFacesCCWOrdered() const
+{
+    return false;
+}
+
+/*!
+    Get the local index of the vertex occupying the n-th position in the
+    anti-clockwise ordered list of faces.
+
+    \param cell is the cell
```
marcocisternino, line 1559: no cell argument in the method prototype. same for face...

andrea-iob:
I've update the documentation and some function/variable names.

marcocisternino (approved):

PR #401, opened 2023-06-07
Third parties

haysam:
THIRPARTIES.md contains a lot of third party components and their licenses to scope with inherited license obligations


edoardolombardi:
> bitpit is not redistributing (nor installing) any of these libraries.
> 
> Also, what if someone downloads bitpit and links it against OpenBLAS, or ATLAS or MKL? Should we include the licenses of all possible BLAS implementations? The same applies to MPI (which seems missing from the list) and to METIS (one can use SCOTCH instead of METIS).
> 
> I mean, we can include the file if it is needed by the GDPR, I'm just wondering how we can include the licenses of all possible third party libraries bitpit can be linked against (bitpit needs the glibc library, should we include also the license of that library?).

I agree. If we don't redistribute libraries with bitpit we don't need third parties licenses. We could update the required dependencies in [INSTALL.md](https://github.com/optimad/bitpit/blob/master/INSTALL.md) if needed (for instance, boost and rapidjson [both optionals] seem missing).

THIRDPARTIES.md
```diff
@@ -0,0 +1,149 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FIT-
+NESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+
+#######################################################################
+RAPIDJASON
```
marcocisternino, line 95: RAPIDJSON!!!

marcocisternino (changes requested):
What about MPI?

andrea-iob:
- Fixed some typos
- Added Open MPI
- Re-formatted the text to make it prettier
- Sorted the software in alphabetic order
- Canceled the CI tests (we can merge this pull request without running the tests)

andrea-iob (approved):

marcocisternino (approved):

PR #404, opened 2023-06-15
containers: allow to pass a constant kernel when setting dynamic synchronization of a PiercedStorage

andrea-iob:
The rationale for this change is that when a storage is registered for synchronization, it won't alter the kernel in a way that affects the usage of the kernel. What happens is that a new PiercedSyncSlave will be added to the list of objects that should be keep in sync with the kernel.

### DIFF
--- a/src/patchkernel/patch_kernel.cpp
+++ b/src/patchkernel/patch_kernel.cpp
@@ -523,7 +523,7 @@ void PatchKernel::initialize()
 	initializeHaloSize(haloSize);
 
 	// Mark patch as partioned
-	if (isPartitioned()) {
+	if (getCommunicator() != MPI_COMM_NULL) {
 		setPartitioningStatus(PARTITIONING_CLEAN);
 	} else {
 		setPartitioningStatus(PARTITIONING_UNSUPPORTED);
--- a/src/patchkernel/patch_kernel.cpp
+++ b/src/patchkernel/patch_kernel.cpp
@@ -1233,7 +1233,7 @@ void PatchKernel::write(VTKWriteMode mode)
 	}
 
 	int vtkVertexCount = 0;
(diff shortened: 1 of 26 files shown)


andrea-iob:
I've fixed a problem with the destruction of the PiercedStorage class, now tests are working again.

src/patchkernel/patch_kernel.cpp
```diff
@@ -523,7 +523,7 @@ void PatchKernel::initialize()
 	initializeHaloSize(haloSize);
 
 	// Mark patch as partioned
-	if (isPartitioned()) {
+	if (getCommunicator() != MPI_COMM_NULL) {
```
edoardolombardi, line 526: Function isCommunicatorSet() could be called, @andrea-iob?

But this doesn't mean that the patch is partitioned, only that the communicator is set. 

The meaning of _partitioned_ patch, from related method isPartitioned(), seems to be that the number of processors is greater than one.

What's the meaning of PARTITIONING_CLEAN status? And when is valid PARTITIONING_UNSUPPORTED? 
A single processor communicator (not null) should be related to a PARTITIONING_UNSUPPORTED patch or always PARTITIONING_CLEAN status?

andrea-iob: PARTITIONING_UNSUPPORTED means that the patch is not able to partition itself, like the VolCartesian patch.

PARTITIONING_CLEAN means that there are no partitioning operation in progress (i.e., we are not in between one of the prepare/alter/finalize methods). 

The correct way to check if the patch is partitioned is isPartitioned(). The partitioning status is exposed, but I've never used it outside bitpit. It may be useful for checking if the patch support partitioning.

edoardolombardi (approved):

PR #405, opened 2023-06-15
POD: miscellanuous fixes

andrea-iob:
Fixes include:
 - usage of smart pointers to avoid manual allocations (Coverity has detected a double free in the code, it was easier to switch to smart pointers rather than fix the double free);
 - data needed by the examples has been updated (it was dumped with an old version of the patch and was not working anymore);
 - one of the example was not working because the POD was trying to set the kernel of a PiercedStorage without first usnetting the existing kernel).
 
Depend on #404.


andrea-iob:
The commits related to this pull request are the latest 4.

edoardolombardi (approved):

PR #409, opened 2023-08-25
containers: add a dedicated function to swap boolean values stored in a PiercedStorage

andrea-iob:
The generic function was triggering a compiler warnings because using the move operator on a temporary object (i.e., on std::vector<bool>::reference) prevents copy elision.

### DIFF
--- a/src/containers/piercedStorage.tpp
+++ b/src/containers/piercedStorage.tpp
@@ -909,10 +909,33 @@ void PiercedStorage<value_t, id_t>::rawErase(std::size_t pos, std::size_t n)
 /**
 * Swaps two elements.
 *
+* This template is used for all types of values with the exception of boolean
+* values.
+*
 * \param pos_first is the position of the first element that will be swapped
 * \param pos_second is the position of the second element that will be swapped
 */
 template<typename value_t, typename id_t>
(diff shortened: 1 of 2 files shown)


src/containers/piercedStorage.tpp
```diff
@@ -913,6 +913,24 @@ void PiercedStorage<value_t, id_t>::rawErase(std::size_t pos, std::size_t n)
...
+{
+    std::size_t firstOffset  = pos_first * m_nFields;
+    std::size_t secondOffset = pos_second * m_nFields;
+    for (std::size_t k = 0; k < m_nFields; ++k) {
+        std::swap(m_fields[firstOffset + k], m_fields[secondOffset + k]);
+    }
+}
+
+/**
+* Swaps two elements.
```
marcocisternino, line 927: "Swaps two Boolean elements", maybe?

marcocisternino (approved):
I cannot see any difference in a bitpit-based application behaviour.
Code checked.

src/containers/piercedStorage.tpp
andrea-iob: I've explicitly stated that the template is used when the container stores boolean values.

PR #410, opened 2023-08-25
IO: add a generic function to flush VTK values 

andrea-iob:
The new function allows to easily flush values on both ASCII and appended files.


marcocisternino (changes requested):
It works, but why isAppendedActive() and isASCIIActive prototypes are in the first commit and their implementation are in the second one?

marcocisternino (approved):

PR #411, opened 2023-08-28
LA: properly setup the preconditioner when solving multiple linear systems

andrea-iob:
From the PETSc user manual: "To solve successive linear systems that have different preconditioner matrices (i.e., the matrix elements and/or the matrix data structure change), the user must call KSPSetOperators() and KSPSolve() for each solve."


ksamouchos (approved):

andrea-iob:
It seems KSPSetUp should be called just once. I refactored the code to make sure only KSPSetOperators is called for each solve (as before, KSPSetOperators is not called if the matrix hasn't changed).

andrea-iob:
Lazy allocation of matrix storage was giving me some troubles, because at the time allocation was performed the information about the estimated number of non-zero element was not available anymore. I reverted the change and only added the functions to handle value and pattern storages.

PR #412, opened 2023-09-10
containers: fix function to set the id of an empty positions of a PiercedKernel

marcocisternino:
When a hole is created the setPosEmptyId made the m_ids vector non-coherent.
This solution is the simplest one, but the less efficient one when the kernel has to be iterated.
This solution solved issues in application using piercedKernel

### DIFF
--- a/src/containers/piercedKernel.tpp
+++ b/src/containers/piercedKernel.tpp
@@ -683,6 +683,7 @@ void PiercedKernel<id_t>::dump() const
 template<typename id_t>
 void PiercedKernel<id_t>::checkIntegrity() const
 {
+    // Check if the elements and their position match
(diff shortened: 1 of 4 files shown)


andrea-iob:
I've did the following changes:
 - checkIntegrirty is now called after every operation performed in the tests;
 - fixed the warnings about size_t/long conversion in the new checks of checkIntegrity;
 - added some comments in checkIntegrirty;
 - updated the commit messages (according to git best practices, the second line of the commit message should be blank);
 
commit

 - the consecutiveness test is now performed only if the kernel is not emtpty (otherwise it will enter in an infinite loop).

andrea-iob (approved):

andrea-iob:
Can we merge this?

PR #413, opened 2023-09-10
cmake: fixes to bitpit installation

andrea-iob:
Pull request #399 introduced some problem with bitpit installation:
 - with some versions of cmake, the cmake files related to bitpit were installed in the wrong directory;
 - headers where installed in the hardcoded "include" directory instead of the path defined by GNUInstallDirs.


marcocisternino (approved):
Code checked. bitpit built and installed to be used in an external project with no problem.

marcocisternino (approved):

marcocisternino:
https://github.com/optimad/bitpit/pull/413/commits/9d108b86492549477257367fda6c710c68657207 is exactly what I've tested.

PR #418, opened 2023-09-21
containers/piercedVector: the storage can always be constructed using a constant kernel

andrea-iob:
After #404 the kernel passed to the constructors of a PiercedStorage can always be constant.


marcocisternino (approved):
I checked the code. It seems that the kernel argument has to be const, the methods called by PiercedStorage constructor and using the kernel argument already treat it as const, they basically are the PiercedStorageSyncSlave constructors.

PR #419, opened 2023-09-21
levelset: fix initialization of segmentation limited-normal validity flag

andrea-iob:
The validity flag should be explicitly set to false, otherwise, if the surface is set multiple times, the container may contain elements wrongly set to true (they were set to true during the evaluation of the levelset with the previous surface).


marcocisternino (approved):
Tested with a parallel case previously giving different results with different number of processes

PR #422, opened 2023-09-28
levelset: rework overall design

andrea-iob:
The main purpose of the rework is to reduce the memory footprint of the levelset. To achieve this goal, the values computed by the levelset are stored in separate caches. Each cache can be enabled separately form the other and can operate in one of the following modes:
 - values can be cached as their are evaluated;
 - values can be pre-computed inside the narrow band; 
 - values can be pre-computed on the whole domain.

When no caches are enabled, the only memory used by the levelset is the memory that each object uses to store its own information (for example the object that evaluates the levelset of a segmentation stored information about the normals of the segmentation).

When no cached are enabled, the evaluation of the levelset is thread safe and this allows to use the levelset object inside parallel OpenMP regions.

This new levelset aims to be 100% compatible with the existing levelset. Some functions are now deprecated, but thay should still work as before.

The pull request is a draft, because more tests should be performed.


andrea-iob:
The final implementation is not 100% compatible with the current version. The names of a couple of classes have changed, but, overall, I think that adapting an existing code to the new implementation should be easy.

andrea-iob:
I fixed a problem with the update of empty objects. Hopefully, this should address the last problems we had with this branch.

marcocisternino (approved):
I tested in a complex geometry case. Differences are very small and results very similar.

PR #425, opened 2023-10-15
patchkernel: empty and serial patches can be partitioned using METIS

andrea-iob:
There is nothing to do when partitioning empty or serial patches, therefore they can be partitioned also using METIS.


marcocisternino (approved):
I approve but the message of the commit should be negative

PR #427, opened 2023-10-18
CG/CG_elem: rework point/plane intersection evaluation

ksamouchos:
This pull request:

1. adds a new function for evaluating the point-line intersection
2. improves the test checking the point-segment intersection
3. micro-optimizes the function evaluating the point-segment intersection (there is one less square root).

This pull-request completes task [IMM-728](https://optimad.youtrack.cloud/issue/IMM-728).

### DIFF
--- a/src/CG/CG_elem.cpp
+++ b/src/CG/CG_elem.cpp
@@ -2531,11 +2531,10 @@ bool intersectPointSegment( array3D const &P, array3D const &P1, array3D const &
 {
     assert( validSegment(P1,P2) );
 
-    if(distancePointSegment(P, P1, P2) > distanceTolerance){
-        return false;
-    }
+    array3D xP     = projectPointSegment( P, P1, P2 );
+    array3D offset = xP - P;
 
-    return true;
+    return (dotProduct(offset, offset) <= distanceTolerance * distanceTolerance);
 }
 
 /*!
--- a/src/CG/CG_elem.cpp
+++ b/src/CG/CG_elem.cpp
@@ -2500,6 +2500,26 @@ bool intersectBoxSphere( array3D const &A0, array3D const &A1, array3D const &ce
 //to levelset }
 
 /*!
+ * checks if points lies on line defined by two given points in 3D
+ * \param[in] P point coordinates
+ * \param[in] Q point on line
+ * \param[in] n line direction
+ * \param[in] distanceTolerance if distance among features exceed this value they are considered as not intersecting
+ * \return if point lies on segment
+ */
+bool intersectPointLine( array3D const &P, array3D const &Q, array3D const &n, const double distanceTolerance)
+{
+    assert( validLine(Q,n) );
+
+    array3D xP;
+    if(distancePointLine(P, Q, n, xP) > distanceTolerance){
+        return false;
+    }
+
+    return true;
+}
+
+/*!
  * checks if points lies on segment in 3D
  * \param[in] P point coordinates
  * \param[in] P1 start point of segment
--- a/test/integration_tests/CG/test_CG_00002.cpp
+++ b/test/integration_tests/CG/test_CG_00002.cpp
@@ -859,6 +859,20 @@ bool testIntersectBoxPolygon(){
 }
 
 /*!
+ * \return true, test intersectPointLine passed
+ */
+bool testIntersectPointLine(){
+    
+    array3D p, n, target;
+    p ={{3,4,5}};
+    n = {{1.4, -5.6, 9.9}};
+    n /= norm2(n);
+    target = p + 1.4453623*n;
+    
+    return intersectPointLine(target, p ,n);
+}
+
+/*!
  * \return true, test intersectPointSegment passed
  */
 bool testIntersectPointSegment(){
@@ -1490,6 +1504,15 @@ int main(int argc, char *argv[])
 
         pass = testIntersectBoxPolygon();
         std::cout << "Testing intersectBoxPolygon...";
+        if (!pass) {
+            std::cout << " Failed" << std::endl;
+            return 1;
+        } else {
+            std::cout << " Passed" << std::endl;
+        }
+
+        pass = testIntersectPointLine();
+        std::cout << "Testing intersectPointLine...";
         if (!pass) {
             std::cout << " Failed" << std::endl;
             return 1;
--- a/test/integration_tests/CG/test_CG_00002.cpp
+++ b/test/integration_tests/CG/test_CG_00002.cpp
@@ -582,6 +582,7 @@ bool testIntersectLinePlane(){
  */
 bool testIntersectSegmentPlane(){
     
+    // Test 1: plane intersects segment
     bool check(true);
     double cc = pow(29,0.5);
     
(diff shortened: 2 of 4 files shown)


src/CG/CG_elem.cpp
```diff
@@ -1876,9 +1876,7 @@ bool intersectSegmentPlane( array3D const &Q1, array3D const &Q2, array3D const
     array3D n = Q2 - Q1;
     n /= norm2(n);
 
-    array3D xP;
-    if ( intersectLinePlane(Q1, n, P2, n2, xP, distanceTolerance) && intersectPointSegment(xP, Q1, Q2, distanceTolerance) ) {
-        P = xP;
+    if ( intersectLinePlane(Q1, n, P2, n2, P, distanceTolerance) && intersectPointSegment(P, Q1, Q2, distanceTolerance) ) {
```
edoardolombardi, line 1879: From documentation: 
> P intersection point if intersect, else unaltered

With this modification it will be altered even if there is no actual intersection between plane and segment, but only between line and plane, @ksamouchos isn't it?

ksamouchos: Correct! I have to change the documentation as well. This change has only to do with personal preferences. So, if you don't agree, I can retrieve the initial behavior of the function.

src/CG/CG_elem.cpp
```diff
@@ -2511,11 +2528,25 @@ bool intersectPointSegment( array3D const &P, array3D const &P1, array3D const &
...
+    if (utils::DoubleFloatingGreater()(t, 0.0, distanceTolerance)
+    && utils::DoubleFloatingLess()(t, 1.0, distanceTolerance)) {
+
+        // Compute projection of P on the segment
+        array3D xP = (1. - t) * P1 + t * P2;
+
+        // Check if P lays on the line defined by the segment
+        if (norm2(P - xP) < distanceTolerance){
+            return true;
+        }
```
andrea-iob, line 2546: In bitpit we have a function that projects a point to a segment, instead of re-writing such functionality you could just use the function we already have:
```
    assert( validSegment(P1,P2) );

    array3D xP     = projectPointSegment( P, P1, P2 );
    array3D offset = xP - P;

    return (dotProduct(offset, offset) < distanceTolerance * distanceTolerance);
```

test/integration_tests/CG/test_CG_00002.cpp
```diff
@@ -1497,7 +1519,16 @@ int main(int argc, char *argv[])
             std::cout << " Passed" << std::endl;
         }
 
-        pass = testIntersectPointSegment();
```
andrea-iob, line 1500: It seems this change removes the point-segment intersection test.

src/CG/CG_elem.cpp
```diff
@@ -2499,6 +2497,25 @@ bool intersectBoxSphere( array3D const &A0, array3D const &A1, array3D const &ce
...
 
+/*!
+ * checks if points lies on line defined by two given points in 3D
+ * \param[in] P point coordinates
+ * \param[in] P1 point belonging to the line
+ * \param[in] P2 point belonging to the line
+ * \param[in] distanceTolerance if distance among features exceed this value they are considered as not intersecting
+ * \return if point lies on segment
+ */
+bool intersectPointLine( array3D const &P, array3D const &P1, array3D const &P2, const double distanceTolerance)
```
andrea-iob, line 2508: The functions that operates on lines should take in input a point on the line and the direction of the line (see other *Line* functions):
```
/*!
 * checks if points lies on line in 3D
 * \param[in] P point coordinates
 * \param[in] Q point on line
 * \param[in] n line direction
 * \param[in] distanceTolerance if distance among features exceed this value they are considered as not intersecting
 * \return if point lies on line
 */
bool intersectPointLine( array3D const &P, const array3D &Q, const array3D &n, const double distanceTolerance)
{
    array3D xP;
    if(distancePointLine(P, Q, n, xP) > distanceTolerance){
        return false;
    }

    return true;
}
```

src/CG/CG_elem.cpp
andrea-iob: This change doesn't seem related to the bug you are fixing. Please move it to another commit.

edoardolombardi: If the change is not really needed for some reason, for compatibility issues on possible applications using this function, personally, I'd keep the original behavior.

ksamouchos: Done!

src/CG/CG_elem.cpp
ksamouchos: I cannot do that. I didn't use function projectPointSegment, because it enforces the projection point to belong into the segment. By using it, I cannot control if the given point actually belongs to the segment.

src/CG/CG_elem.cpp
ksamouchos: Done!

src/CG/CG_elem.cpp
andrea-iob: It doesn't matter, if the point is not on the segment it will not coincide with its projection (neither the real projection nor the projection "moved" on the segment).

ksamouchos: Right! I did the modification.

andrea-iob (approved):
Before merging please change the commit message from "CG/CG_elem: ..." to just "CG: ...". You should be able to that directly from GitHub.

src/CG/CG_elem.cpp
```diff
@@ -2511,11 +2531,10 @@ bool intersectPointSegment( array3D const &P, array3D const &P1, array3D const &
 {
     assert( validSegment(P1,P2) );
 
-    if(distancePointSegment(P, P1, P2) > distanceTolerance){
-        return false;
-    }
+    array3D xP     = projectPointSegment( P, P1, P2 );
```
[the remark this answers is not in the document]
andrea-iob: In the old implementation, this point was considered intersecting the segment:

point_segment_intersection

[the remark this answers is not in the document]
andrea-iob: But you may be right, distancePointSegment seems the function that contains the bug.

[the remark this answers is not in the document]
ksamouchos: I think @edoardolombardi is right, but I still see different (and more correct) results in immerflow with the new implementation. I have to investigate more to see what's going on.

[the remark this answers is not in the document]
andrea-iob: To have an exact match between the two implementations, the updated one should check for "<=" rather than for just "<". However, I don't think this will make any difference for your test case.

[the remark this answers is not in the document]
ksamouchos: I'm studying a case where a plane is almost intersecting a segment. The distance of the projection point P from the segment edge P2 (see above figure) is of order 10^-20. In that case, the order of operations plays a significant role. So, bitpit and an algorithm I've developed to compute intersections were giving me contradictive results. That's why I got this false alarm.

Since bitpit is not wrong, I suggest not to merge this branch and cancel the pull-request. I can continue my work by changing the tolerance used in bitpit to compare distances.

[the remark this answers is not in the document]
andrea-iob: (Please also replace the < with a <= to make the results as close as possible to the previous implementation.)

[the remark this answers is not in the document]
ksamouchos: Thanks! I did what you asked.

edoardolombardi (approved):

PR #431, opened 2023-11-12
discretization: allow to evaluate partial reconstruction weights

andrea-iob:
There are cases in which weights should be evaluated using the coefficients associated with only some of the equations.


marcocisternino (approved):
Tested in application. Results are preserved even in large cases

PR #442, opened 2024-01-01
IO: use Boost library for reading/writing configuration files

andrea-iob:
Using boost allows to drop libxml2 and RapidJOSN dependencies. These two dependencies are tricky to install on Windows, dropping them will make it easy to install bitpit in Windows.


roccoarpa:
Hi, apologies, I prefer to leave the review to the others.

edoardolombardi:
The pull modifications are fine. Anyway, as it is, it's compiling only for Linux, even if its goal is to simplify the Windows compilation. 
Thus, as the pull is related to issue affecting Windows version of the library, it should be rebased on a msvc compliant branch. The current branch rebased on top of https://github.com/optimad/bitpit/tree/general.msvc.compiling.for.AIob can be found in https://github.com/optimad/bitpit/tree/io.configuration.with.boost.msvc, @andrea-iob.

Furthermore, as libxml2 and rapidjson library are no more needed, the download of them through Nuget tool  during project configuration, as done in the mscv compliant branches should be removed. 

The pull should be merged together (or after, in any case rebased on) the complete support of Windows OS. 
Please, consider that the msvc related branch/es need continuous rebases on top of the master branch to be kept aligned and often conflicts happen, mostly in cmake project files.

edoardolombardi:
In my opinion it's not a good practice coding _preparatory_ developments for Windows without testing them on Windows. Indeed, in order to know if their introduction doesn't affect Windows compliant branch, we had to rebase it and test the library compiled on Windows for our purposes. 
The Windows porting should have higher priority than interventions aimed to the optimization of Windows compliancy as this one.
As the branch on Linux and the Windows rebase are ok, I'm approving, but, in order to avoid external rework needed to test the pulls, I think the choice of the prioritization should be reviewed.

edoardolombardi (approved):

PR #443, opened 2024-01-19
Fix voloctree volume mapper

marcocisternino:
In the _recoverPartition method of the voloctree volume mapper, particularly within the algorithm responsible for constructing the list of rank-local mapped mesh octants intended for transmission to the overlapping ranks of the reference mesh, there was an oversight in correctly re-initializing the mapped mesh octant Morton code within the main loop.
Furthermore, a critical check was absent to halt the loop from inserting mapped mesh octants into the list once the mapped mesh rank reaches its final element.


src/voloctree/voloctree_mapper.cpp
```diff
@@ -1237,13 +1237,21 @@ bool VolOctreeMapper::_recoverPartition()
...
         uint64_t reference_last_morton = partitionLDReference[reference_rank];
+        // Initialize idx position in mapped tree to the first mapped
+        // octant inside the overlapping region
         while (morton < reference_first_morton) {
             idx++;
             if (idx == mappedPatch->getTree().getNumOctants()) {
                 break;
             }
             morton = mappedPatch->getTree().getLastDescMorton(idx);
         }
```
andrea-iob, line 1248: Would it make sense to set the morton to an invalid number when idx reaches the end of the octant list?
```
        ...
        while (morton < reference_first_morton) {
            idx++;
            if (idx == mappedPatch->getTree().getNumOctants()) {
                morton = PABLO::INVALID_MORTON;
                break;
            }
            morton = mappedPatch->getTree().getLastDescMorton(idx);
        }
        ...
```
(The new check can then test the Morton number being valid.)

Hopefully this is not needed, but it may may help debugging possible problems in the same function.

edoardolombardi: I think it's clearer in this way. The loop is reaching the final octant of the structure, not an invalid element. The second while is to skip if the index is at the final stage of the octant container, putting an invalid morton could be misunderstanding. 
If the check on num. octants is not clear (in my opinion it is) we could put a boolean variable, identifying the reaching of the end of the octants container and to be checked it before entering in the second while loop, of an old-style, but always used for nested loop, _goto_ statement at the place of the first break.

edoardolombardi (approved):

andrea-iob:
It's not much about the check (we can leave it as it is now), but having the discrepancy ```idx = nOctants``` and ```morton = morton of the (nOctant - 1) octant```.

edoardolombardi:
> It's not much about the check (we can leave it as it is now), but having the discrepancy `idx = nOctants` and `morton = morton of the (nOctant - 1) octant`.

Yes, you're right, for this reason it makes sense. For the check I would keep for clarity the num octants or analogous.

src/voloctree/voloctree_mapper.cpp
```diff
@@ -1232,32 +1232,52 @@ bool VolOctreeMapper::_recoverPartition()
...
     std::map<int, std::vector<long>> list_globalId;
-    uint32_t idx = 0;
-    uint64_t morton = mappedPatch->getTree().getLastDescMorton(idx);
-    for (int reference_rank : toreference_rank[m_rank]) {
+
+    // Initialize idx position in mapped tree to the first mapped
+    // octant inside the overlapping region
+    std::vector<int>::iterator overlap_ref_rank_end_it = toreference_rank[m_rank].end();
+    std::vector<int>::iterator first_overlap_rank_it = overlap_ref_rank_end_it;
+    uint32_t first_overlap_ref_idx = 0;
```
edoardolombardi, line 1240: Why ref? This is not referred to mapped patch?

edoardolombardi (approved):

andrea-iob (approved):

src/voloctree/voloctree_mapper.cpp
marcocisternino: Because it is an error. It should be called first_overlap_mapped_idx. I rename it.

marcocisternino:
I renamed "first_overlap_ref_idx" in "first_overlap_map_idx".
This remove the confusion about the meaning of the variable.

marcocisternino:
> It's not much about the check (we can leave it as it is now), but having the discrepancy `idx = nOctants` and `morton = morton of the (nOctant - 1) octant`.

It should be cleaner now, do you agree?

marcocisternino:
I put this pull in draft. Further tests show problems in the splitted algorithm. At [b59ef91](https://github.com/optimad/bitpit/pull/443/commits/b59ef9199d6f76cef755352678b71335b04eb6f1) the algorithm works fine.

src/voloctree/voloctree_mapper.cpp
```diff
@@ -1232,32 +1232,52 @@ bool VolOctreeMapper::_recoverPartition()
...
-    uint64_t morton = mappedPatch->getTree().getLastDescMorton(idx);
-    for (int reference_rank : toreference_rank[m_rank]) {
+
+    // Initialize idx position in mapped tree to the first mapped
+    // octant inside the overlapping region
+    std::vector<int>::iterator overlap_ref_rank_end_it = toreference_rank[m_rank].end();
+    std::vector<int>::iterator first_overlap_rank_it = overlap_ref_rank_end_it;
+    uint32_t first_overlap_map_idx = 0;
+    for (std::vector<int>::iterator ref_rank_it = toreference_rank[m_rank].begin(); ref_rank_it != overlap_ref_rank_end_it; ++ref_rank_it) {
+        first_overlap_map_idx = 0;
```
edoardolombardi, line 1242: [see @andrea-iob comment](https://github.com/optimad/bitpit/commit/c279845277ac29be204ba9dc39e7827066aa7450#r138043356)

andrea-iob:
I don't see the latest changes, can you check if you have uploaded them?

marcocisternino:
As much as I could test it, it seems to work now.

edoardolombardi (approved):

src/voloctree/voloctree_mapper.cpp
```diff
@@ -1232,32 +1232,51 @@ bool VolOctreeMapper::_recoverPartition()
...
+        // Fill partitioning info structure with mapped octants in the
+        // overlapping region
+        if (first_overlap_map_idx < mappedPatch->getTree().getNumOctants() ) {
+            uint32_t idx = first_overlap_map_idx;
+            for (std::vector<int>::iterator ref_rank_it = overlap_ref_rank_begin_it; ref_rank_it != overlap_ref_rank_end_it; ++ref_rank_it) {
+                int reference_rank = *ref_rank_it;
+                uint64_t reference_last_morton = partitionLDReference[reference_rank];
+                uint64_t morton = mappedPatch->getTree().getMorton(idx);
+                while (morton < reference_last_morton) {
+                    Octant oct = *mappedPatch->getTree().getOctant(idx);
```
andrea-iob, line 1262: Do you really want to copy the octant?

andrea-iob (approved):

src/voloctree/voloctree_mapper.cpp
marcocisternino: It depends on who is "you".
We could avoid some copies, locally reworking the method. Even if it was the original algorithm that copies octants, I added a commit to avoid local octant copies. Further changes in this direction should addressed in another pull request.
Finally, I rebased on current master

PR #446, opened 2024-02-09
patchkernel: find all the closest cells to a given point

marcocisternino:
This pull request adds methods to SurfaceSkdTree to find all the closest cells to a given point. The aim is to provide external applications with the the chance to select the the closest cell using its criteria. Existing methods select the closest cell using a specific criterium: the cell more "aligned" with the line that connect the specified point and its projection onto the cell will be chosen.

### DIFF
--- a/src/patchkernel/surface_skd_tree.cpp
+++ b/src/patchkernel/surface_skd_tree.cpp
@@ -369,111 +369,20 @@ long SurfaceSkdTree::findPointClosestCell(const std::array<double, 3> &point, do
         return 0;
     }
 
-    // Initialize a distance estimate
-    //
-    // The real distance will be lesser than or equal to the estimate.
-    //
-    // Care must be taken to avoid overflow when performing the multiplication.
-    double squaredMaxDistance;
-    if (maxDistance <= 1. || maxDistance < std::numeric_limits<double>::max() / maxDistance) {
-        squaredMaxDistance = maxDistance * maxDistance;
-    } else {
-        squaredMaxDistance = std::numeric_limits<double>::max();
-    }
-
-    double squareDistanceEstimate = std::min(root.evalPointMaxSquareDistance(point), squaredMaxDistance);
-
-    // Get a list of candidates nodes
-    //
-    // If threads safe lookups are not needed, some temporary data structures
-    // are declared as member of the class to avoid their reallocation every
-    // time the function is called.
-    //
-    // First, we gather all the candidates and then we evaluate the distance
-    // of each candidate. Since distance estimate is constantly updated when
-    // new nodes are processed, the final estimate may be smaller than the
-    // minimum distance of some candidates. Processing the candidates after
-    // scanning all the tree, allows to discard some of them without the need
-    // of evaluating the exact distance.
-    std::unique_ptr<std::vector<std::size_t>> privateNodeStack;
-    std::unique_ptr<std::vector<std::size_t>> privateCandidateIds;
-    std::unique_ptr<std::vector<double>> privateCandidateMinDistances;
-
-    std::vector<std::size_t> *nodeStack;
-    std::vector<std::size_t> *candidateIds;
-    std::vector<double> *candidateMinDistances;
-    if (areLookupsThreadSafe()) {
-        privateNodeStack = std::unique_ptr<std::vector<std::size_t>>(new std::vector<std::size_t>());
-        privateCandidateIds = std::unique_ptr<std::vector<std::size_t>>(new std::vector<std::size_t>());
-        privateCandidateMinDistances = std::unique_ptr<std::vector<double>>(new std::vector<double>());
-
-        nodeStack = privateNodeStack.get();
-        candidateIds = privateCandidateIds.get();
-        candidateMinDistances = privateCandidateMinDistances.get();
-    } else {
-        m_nodeStack.clear();
-        m_candidateIds.clear();
-        m_candidateMinDistances.clear();
-
-        nodeStack = &m_nodeStack;
-        candidateIds = &m_candidateIds;
-        candidateMinDistances = &m_candidateMinDistances;
-    }
-
-    nodeStack->push_back(rootId);
-    while (!nodeStack->empty()) {
-        std::size_t nodeId = nodeStack->back();
-        const SkdNode &node = m_nodes[nodeId];
-        nodeStack->pop_back();
-
-        // Do not consider nodes with a minimum distance greater than
-        // the distance estimate
-        double nodeMinSquareDistance = node.evalPointMinSquareDistance(point);
-        if (utils::DoubleFloatingGreater()(nodeMinSquareDistance, squareDistanceEstimate, tolerance, tolerance)) {
-            continue;
-        }
-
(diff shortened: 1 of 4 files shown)


src/patchkernel/surface_skd_tree.cpp
```diff
@@ -581,6 +490,226 @@ long SurfaceSkdTree::findPointClosestCell(int nPoints, const std::array<double,
...
+* \param[in] maxDistance all cells whose distance is greater than
+* this parameters will not be considered for the evaluation of the
+* distance
+* \param[out] distanceEstimate on output it will contain the estimate
+* of the distance between the point and closest cell. If all cells
+* contained in the tree are farther than the maximum distance, the
+* argument will be set to the maximum representable distance
+*/
+void SurfaceSkdTree::findPointClosestCandidates(const std::array<double, 3> &point, double maxDistance,
+        double *distanceEstimate, Candidates &candidates) const
```
andrea-iob, line 613: Pass it by pointer.

src/patchkernel/surface_skd_tree.hpp
```diff
@@ -68,6 +72,15 @@ class SurfaceSkdTree : public PatchSkdTree {
     mutable std::vector<std::size_t> m_candidateIds;
     mutable std::vector<double> m_candidateMinDistances;
 
+    struct Candidates {
+        std::vector<std::size_t> *candidateIds = nullptr;
+        std::vector<double> *candidateMinDistances = nullptr;
+        std::vector<std::size_t> privateCandidateIds;
+        std::vector<double> privateCandidateMinDistances;
```
andrea-iob, line 79: These should be a unique_ptr of a vector to avoid the allocation. Also, rename it to candidateTdsStorage, candidateMinDistancesStorage.

src/patchkernel/surface_skd_tree.cpp
andrea-iob, line 613: Since the Candidates already contain information about the distance of the candidates, we can move the distance estimate into the structure. We can call it maximumDistance.

src/patchkernel/surface_skd_tree.hpp
andrea-iob: The structure already contains the name candidates, we can remove it from its memebers.

src/patchkernel/surface_skd_tree.cpp
andrea-iob: Or maxDistance to be coherent with the other member of the distance.

src/patchkernel/surface_skd_tree.hpp
andrea-iob: Move the declaration before the members.

src/patchkernel/surface_skd_tree.cpp
```diff
@@ -581,6 +490,226 @@ long SurfaceSkdTree::findPointClosestCell(int nPoints, const std::array<double,
     return nDistanceEvaluations;
 }
 
+
+/*!
+* Given the specified point find the closest cells contained in the tree and
+* evaluate the distance between that cells and the given point. All the cells
+* at the same distance (using patch tolerance) are returned.
```
edoardolombardi, line 497: At the same distance? Or all the cells at a distance <= than the max distance?

src/patchkernel/surface_skd_tree.cpp
```diff
@@ -581,6 +490,226 @@ long SurfaceSkdTree::findPointClosestCell(int nPoints, const std::array<double,
...
+*/
+long SurfaceSkdTree::findPointClosestCells(const std::array<double, 3> &point, std::vector<long> &ids, double *distance) const
+{
+    return findPointClosestCells(point, std::numeric_limits<double>::max(), false, ids, distance);
+}
+
+/*!
+* Given the specified point find the closest cells contained in the tree and
+* evaluate the distance between that cells and the given point. All the cells
+* at the same distance (using patch tolerance) are returned.
```
edoardolombardi, line 516: At the same distance? Or all the cells at a distance <= than the max distance?

src/patchkernel/surface_skd_tree.cpp
```diff
@@ -581,6 +490,226 @@ long SurfaceSkdTree::findPointClosestCell(int nPoints, const std::array<double,
...
+long SurfaceSkdTree::findPointClosestCells(const std::array<double, 3> &point, double maxDistance,
+                                          std::vector<long> &ids, double *distance) const
+{
+    return findPointClosestCells(point, maxDistance, false, ids, distance);
+}
+
+/*!
+* Given the specified point find the closest cells contained in the tree and
+* evaluate the distance between that cells and the given point. All the cells
+* at the same distance (using patch tolerance) are returned.
```
edoardolombardi, line 539: At the same distance? Or all the cells at a distance <= than the max distance?

src/patchkernel/surface_skd_tree.cpp
edoardolombardi: I think that this method is returning the whole geometry, what's the utility?

edoardolombardi (approved):

src/patchkernel/surface_skd_tree.cpp
```diff
@@ -581,6 +497,220 @@ long SurfaceSkdTree::findPointClosestCell(int nPoints, const std::array<double,
...
+    // time the function is called.
+    ClosestCellCandidates *closestCellCandidates = nullptr;
+    std::unique_ptr<ClosestCellCandidates> privateStorage = nullptr;
+    if (areLookupsThreadSafe()) {
+        privateStorage = std::unique_ptr<ClosestCellCandidates>(new ClosestCellCandidates());
+        closestCellCandidates = privateStorage.get();
+    } else {
+        closestCellCandidates = &m_closestCellCandidates;
+    }
+    closestCellCandidates->maxDistance = *distance;
```
andrea-iob, line 593: Is necessary to initialize maxDistance?

src/patchkernel/surface_skd_tree.cpp
```diff
@@ -581,6 +497,220 @@ long SurfaceSkdTree::findPointClosestCell(int nPoints, const std::array<double,
...
+    findPointClosestCandidates(point, maxDistance, closestCellCandidates);
+
+    // Process the candidates and find the closest cells
```
andrea-iob, line 596: To make evaluation of the distance clearer, please move the initialization if the distance here:
```
*distance = closestCellCandidates->maxDistance;

long nDistanceEvaluations = 0;
for () {
  ... 
}
```
with the code inside the for cycle using distance instead of closestCellCandidates->maxDistance.

andrea-iob (approved):

andrea-iob:
Can you confirm that the changes didn't slow-down the function?

src/patchkernel/surface_skd_tree.cpp
marcocisternino: No it is not

marcocisternino:
> Can you confirm that the changes didn't slow-down the function?

integration levelset tests run sometimes faster and sometimes slower than the current master, therefore, I can statistically confirm no changes in performance.

PR #452, opened 2024-02-28
PABLO: rework the function that evaluates the octant owner of a point

andrea-iob:
If the owner of the point is an internal octant, decrementing the Morton upper bound by one will give us the requested octant. If the owner of the point is a ghost octant, we need to test all the octants before the Morton upper bound.


marcocisternino (approved):
code checked and tested using voloctree mesh mapper

PR #454, opened 2024-03-12
levelset: fix dump/restore of caches 

andrea-iob:
The index associated with the caches should be preserved during the dump/restore procedure.


marcocisternino (approved):
Naca test case dump/restore is now working.
Other test cases (3D sphere) were working without this fix and they still work.

PR #458, opened 2024-03-18
external: move CBLAS header in the correct directory 

andrea-iob:
The header belongs to the CBLAS directory.


marcocisternino (approved):
library and application configuration and build tested.
application tests pass

PR #459, opened 2024-03-18
Bump requested C++ standard version to C++17

andrea-iob:
It's 2024, C++17 seems the bare minimum.


edoardolombardi (approved):

marcocisternino:
I added:
- a binary utils write for non-const char*
- c++17 requirement for tests and examples

PR #460, opened 2024-04-02
LA: set block size information even when a flat internal storage is used

andrea-iob:
When the argument "flatten" is set to true, block size will not be taken into account when allocating the internal storage of the system matrix. Even when the internal storage is flat, block size information are available and can be used by the solver to speed up the solution of the system. However, since the internal storage doesn't take blocks into account, some low level operations (e.g., matrix-matrix multiplications) cannot use block information. Some algorithms for the solution of the system (e.g., mutligrid) requires a flat storage.


marcocisternino (approved):
code checked and tested. No performance tests have been performed

PR #461, opened 2024-04-12
levelset: surface smoothing

ksamouchos:
This pull request introduces a new tool which smooths out a discretized surface (polygon or polyhedral). The smoothed surface is continuous and is equipped with a continuous normal. Also, it passes form the discretized geometry's vertices and respects the normal vectors corresponding to each vertex. The new level set computes the projection point and normal based on the new continuous surface.

This pull request completes task [IMM-770](https://optimad.youtrack.cloud/issue/IMM-770).


src/levelset/levelSetCommon.hpp
```diff
@@ -177,6 +177,15 @@ enum class LevelSetField{
     UNDEFINED
 };
 
+/*!
+ * @ingroup levelsetEnums
+ * Enum class containing the possible order of surface smooting
+ */
+enum class LevelSetSurfaceSmoothing{
+    LOW_ORDER,                      /**< level set high order surface smoothing */
+    HIGH_ORDER                      /**< level set low order surface smoothing */
```
andrea-iob, line 186: Add the HIGH_ORDER flag when high order smoothing is introduced.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -298,59 +298,291 @@ std::array<double, 3> LevelSetSegmentationSurfaceInfo::evalNormal(const std::arr
...
+ *
+ * @param[in] point are the coordinates of point
+ * @param[in] segmentItr is an iterator pointing to the closest segment
  * @param[out] lambda on output will contain the barycentric coordinates of the projection point
- * @return The coordinates of the projection point.
  */
-std::array<double, 3> LevelSetSegmentationSurfaceInfo::evalProjection(const std::array<double, 3> &point,
-                                                                      const SegmentConstIterator &segmentItr,
-                                                                      double *lambda) const
+void LevelSetSegmentationSurfaceInfo::evalBarycentricCoordinates(const std::array<double, 3> &point,
```
andrea-iob, line 1159: Can we move evalBarycentricCoordinates in SurfaceKernel (and isnce we are at it, also in LineKernel)?

src/levelset/levelSetSegmentationObject.hpp
```diff
@@ -97,6 +97,15 @@ class LevelSetSegmentationSurfaceInfo {
...
+    void evalLowOrderProjectionOnTriangle(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionPoint, std::array<double, 3> *projectionNormal) const;
+    void evalLowOrderProjectionOnPolygon(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionPoint, std::array<double, 3> *projectionNormal) const;
+    void evalLowOrderProjection(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionPoint, std::array<double, 3> *projectionNormal) const;
+
+    void evalBarycentricCoordinates(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, double *lambda) const;
+    void evalProjection(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionPoint, std::array<double, 3> *projectionNormal) const;
+    void evalProjection(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionPoint) const;
```
andrea-iob: Please add this two functions in the commit that changes evalProjection to evaluate both projection point and projection normal.

src/CG/CG.hpp
```diff
@@ -100,6 +100,7 @@ int convertBarycentricToFlagPolygon( std::size_t, double const *, double toleran
 void computeGeneralizedBarycentric( array3D const &, std::vector<array3D> const &, std::vector<double> &);
 void computeGeneralizedBarycentric( array3D const &, std::size_t, array3D const *, std::vector<double> &);
 void computeGeneralizedBarycentric( array3D const &, std::size_t, array3D const *, double *);
+void computeGeneralizedBarycentricDerivative( array3D const &, std::size_t, array3D const *, double *, std::array<double, 3> *);
```
andrea-iob, line 103: I don't like duplicating the code for the evaluation of the barycentric coordinates. It would be better to either modify the existing computeGeneralizedBarycentric functions to evaluate both barycentric coordinates and their derivative, or add a new set of functions to evaluate the barycentric coordinates derivative.

src/CG/CG_elem.cpp
```diff
@@ -2989,6 +3058,36 @@ double areaTriangle( array3D const &a, array3D const &b, array3D const &c)
...
+    array3D d1_c = {0.0, 1.0, 0.0};
+    array3D d2_c = {0.0, 0.0, 1.0};
+
+    array3D vec = crossProduct(b-a,c-a);
+
+    array3D d0_vec = crossProduct(b-a,d0_c-a);
+    array3D d1_vec = crossProduct(b-a,d1_c-a);
+    array3D d2_vec = crossProduct(b-a,d2_c-a);
+
+    vec /= norm2(vec);
```
andrea-iob, line 3081: Move this line near the evaluation of vec.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+    const Cell &segment = *segmentItr;
+
+    // Get segment vertices
+    ConstProxyVector<long> segmentVertexIds = segment.getVertexIds();
+
+    std::array<double,3> point0 = m_surface->getVertexCoords(segmentVertexIds[0]);
+    std::array<double,3> point1 = m_surface->getVertexCoords(segmentVertexIds[1]);
+
+    // Get projection point on segment
+    int nSegmentVertices = segment.getVertexCount();
```
andrea-iob, line 377: int nSegmentVertices = segmentVertexIds.size();

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+    BITPIT_CREATE_WORKSPACE(t, double, nSegmentVertices, ReferenceElementInfo::MAX_ELEM_VERTICES);
```
andrea-iob, line 378: This code is specific for the line element, ```t``` can be declared as an array of size ReferenceElementInfo::MAX_ELEM_VERTICES.

andrea-iob: As a side note, we should modify ReferenceElementInfo to allow accessing information on reference elements at compile time:
```
	const ReferenceElementInfo &referenceSegment = ReferenceElementInfo::getInfo(ElementType::LINE);
	std::array<double, referenceSegment.nVertices> t;
```

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+ * @param[out] projectionNormal The coordinates of the norrmal to the surface vector on the surface
+ * projection point.
+ */
+void LevelSetSegmentationSurfaceInfo::evalHighOrderProjectionOnTriangle(const std::array<double,3> &point,
+                                                                        const SegmentConstIterator &segmentItr,
+                                                                        std::array<double, 3> *projectionPoint,
+                                                                        std::array<double, 3> *projectionNormal) const
+{
+    // Get segment
+    const Cell &segment = *segmentItr;
```
andrea-iob, line 453: assert(...);

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+
+    // Get segment vertices
+    ConstProxyVector<long> segmentVertexIds = segment.getVertexIds();
+
+    std::array<double,3> point0 = m_surface->getVertexCoords(segmentVertexIds[0]);
```
andrea-iob, line 458: Constant reference.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+    std::array<double,3> point1 = m_surface->getVertexCoords(segmentVertexIds[1]);
+    std::array<double,3> point2 = m_surface->getVertexCoords(segmentVertexIds[2]);
+
+    // Get projection point on segment
+    int nSegmentVertices = segment.getVertexCount();
```
andrea-iob, line 463: segmentVertexIds.size()

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+    BITPIT_CREATE_WORKSPACE(tau, double, nSegmentVertices, ReferenceElementInfo::MAX_ELEM_VERTICES);
```
andrea-iob, line 464: Same as before.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+        normal_s = facetNormal;
+    } else {
+        normal_s /= distance;
+        if (dotProduct(normal_s, facetNormal) < 0.0) {
+            normal_s *= -1.0;
+        }
+    }
+
+    // Compute projection point and normal at each edge "e"
+    std::array< std::array<double, 3>, 3 > edgePoint_s;
```
andrea-iob, line 494: Remove the spaces after/before </>.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+    } else {
+        normal_s /= distance;
+        if (dotProduct(normal_s, facetNormal) < 0.0) {
+            normal_s *= -1.0;
+        }
+    }
+
+    // Compute projection point and normal at each edge "e"
+    int nSegmentEdges = nSegmentVertices;
+    BITPIT_CREATE_WORKSPACE(edgePoint_s, std::array<double BITPIT_COMMA 3>, nSegmentEdges, ReferenceElementInfo::MAX_ELEM_VERTICES);
```
andrea-iob, line 776: MAX_ELEM_EDGES

src/levelset/levelSetSegmentationObject.hpp
```diff
@@ -78,6 +78,11 @@ class LevelSetSegmentationSurfaceInfo {
 
     std::array<double, 3> evalNormal(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr) const;
 
+    void evalBarycentricCoordinates(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, double *lambda) const;
+    void evalProjection(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionPoint, std::array<double, 3> *projectionNormal) const;
+    void evalProjection(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionPoint) const;
+    std::array<double, 3> evalProjection(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, double *lambda) const;
```
andrea-iob, line 85: The API is now heterogeneous, can we the evalProjection functions more consistent? For example:
```
std::array<double, 3> evalProjection(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr) const;
std::array<double, 3> evalProjection(const std::array<double, 3> &point, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionNormal) const;
std::array<double, 3> evalProjection(const std::array<double, 3> &lambda, const SegmentConstIterator &segmentItr) const;
std::array<double, 3> evalProjection(const std::array<double, 3> &lambda, const SegmentConstIterator &segmentItr, std::array<double, 3> *projectionNormal) const;
```

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -2575,7 +2575,6 @@ LevelSetCellLocation LevelSetSegmentationObject::fillCellGeometricNarrowBandLoca
     } else if (cellUnsigendValue <= m_narrowBandSize) {
         cellLocation = LevelSetCellLocation::NARROW_BAND_DISTANCE;
     }
-    assert(cellLocation != LevelSetCellLocation::UNKNOWN);
```
andrea-iob, line 1577: This means that there will be cells that are inside the narrow band but tagged as LevelSetCellLocation::UNKNOWN? I don't think this will work.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -2872,19 +2885,34 @@ std::array<double,3> LevelSetSegmentationObject::_evalGradient(const std::array<
...
     // Evaluate levelset gradient
-    std::array<double,3> gradient = m_surfaceInfo->evalDistanceVector(point, supportItr) / distance;
+    if (signedLevelSet) {
+       std::array<double, 3> pseudoNormal = m_surfaceInfo->evalPseudoNormal(supportItr, lambda);
+       double normalGradient = dotProduct(gradient, pseudoNormal);
+       if (utils::DoubleFloatingEqual()(normalGradient, 0., distanceTolerance, distanceTolerance)) {
+            throw std::runtime_error("Unable to evaluate gradient: the point lies on the normal plane!");
+       }
+       gradient *= sign(normalGradient);
+    }
```
andrea-iob, line 2915: Unless this optimization yields a significant speed-up, I don't like duplicating the code that evaluate the distance in the function that evaluates the gradient. Even if the speed-up is significant, the function may need some rework because it seems it will give wrong results if the point lies on the normal plane and signedLevelSet is false.

src/levelset/levelSetCommon.hpp
andrea-iob: You can remove the "level set" text, just ```//! High order surface smoothing```. Also, the comments are swapped.

andrea-iob:
The code is failing to compile on windows because it is using M_PI. There is no such constant on windows, you should use BITPIT_PI/BITPIT_PI_2/BITPIT_PI_4. 

M_PI is defined only if _USE_MATH_DEFINES is set, however this macro should be set before including cmath. If an external library includes cmath without defining _USE_MATH_DEFINES, then M_PI will not be available. To avoid potential problems, we have decided to create out custom pi constant.

ksamouchos:
Branch update:
- I've corrected the surface smoothing on polygonal segments
- I've added a levelset test case to control the surface smoothing over polygonal segments
- I've made the correction proposed by Andrea
- I've rebased the branch on the current master

src/levelset/levelSetCommon.hpp
ksamouchos: I removed the "level set" text. I don't understand your first comment. Could you please rephrase it?

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done

src/levelset/levelSetSegmentationObject.hpp
ksamouchos: Done

src/CG/CG.hpp
ksamouchos: I deleted completely the function. In the new version I avoid the usage of barycentric coordinates because I changed the algorithm of creating the surface over polygonal segments. Now the resulted surface has smoother behavior.

src/CG/CG_elem.cpp
ksamouchos: I deleted the function. It's not used anymore.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+ * @param[out] projectionNormal The coordinates of the norrmal to the surface vector on the surface
+ * projection point.
+ */
+void LevelSetSegmentationSurfaceInfo::evalHighOrderProjectionOnLine(const std::array<double,3> &point,
+                                                                    const SegmentConstIterator &segmentItr,
+                                                                    std::array<double, 3> *projectionPoint,
+                                                                    std::array<double, 3> *projectionNormal) const
+{
+    // Get segment
+    const Cell &segment = *segmentItr;
```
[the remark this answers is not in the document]
ksamouchos: Done

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+
+    // Get segment vertices
+    ConstProxyVector<long> segmentVertexIds = segment.getVertexIds();
+
+    std::array<double,3> point0 = m_surface->getVertexCoords(segmentVertexIds[0]);
```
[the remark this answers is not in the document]
ksamouchos: Done everywhere

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: I deleted the usage of BITPIT_CREATE_WORKSPACE for lines and triangles.

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done in all similar cases

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -331,6 +331,671 @@ void LevelSetSegmentationSurfaceInfo::evalProjectionOnVertex(const std::array<do
...
+    for (int e = 0; e < nSegmentEdges; e ++) {
+        // Get edge's local and global node ids
+        ConstProxyVector<int> edgeLocalVertexIds = segment.getFaceLocalVertexIds(e);
+        int localId0 = edgeLocalVertexIds[0]; 
+        int localId1 = edgeLocalVertexIds[1]; 
+        int id0      = segment.getVertexId(localId0);
+        int id1      = segment.getVertexId(localId1);
+
+        // Get vertex information
+        std::array<double,3> node0 = m_surface->getVertexCoords(id0);
```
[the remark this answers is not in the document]
ksamouchos: Done

src/levelset/levelSetSegmentationObject.hpp
ksamouchos: The API is heterogeneous because the two "void evalProjection(...)" functions return both the projection point and the normal. I don't like to return the projection point as a return value and the normal as a reference argument. I could change the "std::array<double, 3> evalProjection(...)" to be of type void, but it's a public member and I prefer to keep it as it is.

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: I did the optimization in function evalDistance and I kept the initial version of the gradient evaluation function.

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: The cell location can be unknown because it's evaluated later. It happens in one of the new tests I've created. The removal of the assert is safe because whenever this function is used (fillCellGeometricNarrowBandLocationCache), the scenario of returning an unknown location is always considered into account.

ksamouchos:
> The code is failing to compile on windows because it's using M_PI. There is no such constant on windows, you should use BITPIT_PI/BITPIT_PI_2/BITPIT_PI_4.
> 
> M_PI is defined only if _USE_MATH_DEFINES is set, however this macro should be set before including cmath. If an external library includes cmath without defining _USE_MATH_DEFINES, then M_PI will not be available. To avoid potential problems, we have decided to create our own pi constant.

I didn't know it. Thank you for mentioning it. Now I'm using the BITPI_PI variable.

ksamouchos:
I changed function
```
void setSurface(const SurfUnstructured *surface, double featureAngle, LevelSetSurfaceSmoothing surfaceSmoothing, bool force = false);
```
to allow the user to modify the surface smoothing order of object LevelSetSegmentationObject after its creation.

src/levelset/levelSetCommon.hpp
andrea-iob: > I removed the "level set" text. I don't understand your first comment. Could you please rephrase it?

You have added the HIGH_ORDER constant when the code doesn't yet support high order surface smoothing. The HIGH_ORDER constant should be introduced at the same time as the high order smoothing.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -1796,7 +1797,26 @@ int LevelSetSegmentationBaseObject::evalPart(const std::array<double,3> &point)
...
+/*!
+ * Evaluate the normal of the surface at the segment closest to the specified point.
+ *
+ * \param point are the coordinates of the point
+ * \param signedLevelSet controls if signed levelset function will be used
+ * \result The normal of the surface at the segment closest to the specified point.
+ */
+std::array<double,3> LevelSetSegmentationBaseObject::_evalNormal(const std::array<double,3> &point, bool signedLevelSet) const
+{
+    return evalNormal(point, signedLevelSet);
```
andrea-iob, line 1850: It should be the other way around: the evalNormal function should call the _evalNormal function. Functions that start with a "_" are virtual functions that the derive class should be able to override.

```
void BaseClass::function()
{
    <code common for all classes>

    _function()   <---- Code specific for the derive class

    <code common for all classes> 
}
```

If evalNormal doesn't need to be overridden, just delete the _evalNormal function and do all the work in evalNormal.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -1796,7 +1797,26 @@ int LevelSetSegmentationBaseObject::evalPart(const std::array<double,3> &point)
  */
 std::array<double,3> LevelSetSegmentationBaseObject::evalNormal(const std::array<double,3> &point, bool signedLevelSet) const
 {
-    return _evalNormal(point, signedLevelSet);
+    // Project the point on the surface and evaluate the point-projection vector
+    std::array<double, 3> projectionPoint;
+    std::array<double, 3> projectionNormal;
+    evalProjection(point, signedLevelSet, &projectionPoint, &projectionNormal);
+
+    BITPIT_UNUSED(projectionPoint);
```
andrea-iob, line 1836: There is no need to mark projectionPoint as unused. The macro is only needed to silence compiler workings related to unused function arguments.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -2686,7 +2706,13 @@ std::array<double,3> LevelSetSegmentationObject::_evalCellNormal(long id, bool s
     long support = evalCellSupport(id);
     std::array<double,3> centroid = m_kernel->computeCellCentroid(id);
 
-    return _evalNormal(centroid, support, signedLevelSet);
+    std::array<double, 3> projectionPoint;
+    std::array<double, 3> projectionNormal;
+    _evalProjection(centroid, support, signedLevelSet, &projectionPoint, &projectionNormal);
+
+    BITPIT_UNUSED(projectionPoint);
```
andrea-iob, line 2744: There is no need to mark projectionPoint as unused.

src/levelset/levelSetSegmentationObject.cpp
andrea-iob: One of the places the function is used is here:
```
void LevelSetSegmentationObject::fillCartesianCellZoneCache()
{
        ...

        // Fill location cache for cells geometrically inside the narrow band
        LevelSetCellLocation cellLocation = fillCellGeometricNarrowBandLocationCache(cellId);

        // Track intersected cells
        if (cellLocation == LevelSetCellLocation::NARROW_BAND_INTERSECTED) {
            intersectedCellIds.push_back(cellId);
        }

        // Add unprocessed neighbours of cells inside the narrow band to the process list
        if (cellLocation != LevelSetCellLocation::UNKNOWN) {
            auto neighProcessor = [&processList, &alreadyProcessed](long neighId, int layer) {
                BITPIT_UNUSED(layer);

        ...
}

In this case cellLocation can be UNKNOWN and the meaning is "I don't know where the cell is, but is certainly outside the narrowband band". If the function returns UNKNOWN also for cells inside the narrow band (but that cannot be identified as such at that stage), which portion of the code will add the neighbours of those cells (cells inside the narrow band but whose location is UNKNOWN) to the process list?

File: src/levelset/levelSetSegmentationObject.cpp (hunk 20 above)
Comment (andrea-iob) [reply src/levelset/levelSetSegmentationObject.cpp:1850]: Or maybe keep only the \_evalNormal function in LevelSetSegmentationBaseObject and add there the new implementation.

**Comment by ksamouchos** [comment]
Rebased to the current master

File: src/levelset/levelSetCommon.hpp (hunk 1 above)
Comment (ksamouchos) [reply src/levelset/levelSetCommon.hpp]: Done

File: src/levelset/levelSetSegmentationObject.cpp (hunk 15 above)
Comment (ksamouchos) [reply src/levelset/levelSetSegmentationObject.cpp:1577]: Answered bellow

File: src/levelset/levelSetSegmentationObject.cpp (hunk 20 above)
Comment (ksamouchos) [reply src/levelset/levelSetSegmentationObject.cpp]: Done

File: src/levelset/levelSetSegmentationObject.cpp (hunk 22 above)
Comment (ksamouchos) [reply src/levelset/levelSetSegmentationObject.cpp]: I deleted it. I know it's not needed, but I thought that in this way the code is clearer.

File: src/patchkernel/surface_kernel.hpp [hunk 23]
```diff
@@ -69,7 +69,10 @@ class SurfaceKernel : public PatchKernel {
     std::array<double, 3> evalLimitedVertexNormal(long, int, std::size_t, const long *, double ) const;
     virtual void evalVertexNormals(long id, int vertex, std::size_t nVertexNeighs, const long *vertexNeighs, double limit,
                                    std::array<double, 3> *unlimitedNormal, std::array<double, 3> *limitedNormal) const;
+    virtual void evalEdgeNormals(long id, int edge, double limit, std::array<double, 3> *unlimitedNormal,
+                                 std::array<double, 3> *limitedNormal) const;
     double evalCellSize(long id) const override;
+    void evalBarycentricCoordinates(long id, const std::array<double, 3> &point, double *lambda) const;
```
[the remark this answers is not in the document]
ksamouchos: It's needed. This function is still in the surface kernel class. I just added it to the line kernel as well.

src/patchkernel/line_kernel.cpp
```diff
@@ -22,6 +22,7 @@
  *
 \*---------------------------------------------------------------------------*/
 
+#include "CG.hpp"
```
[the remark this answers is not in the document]
ksamouchos: Done

src/patchkernel/surface_kernel.cpp
```diff
@@ -891,6 +948,93 @@ void SurfaceKernel::evalVertexNormals(long id, int vertex, std::size_t nVertexNe
     }
 }
 
+/*!
+ * Evaluate unlimited and limited normal unit vectors at the specified local
+ * edge.
+ *
+ * Edge normal is computed as the arithmetic * average of the normals to each
```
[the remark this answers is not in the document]
ksamouchos: Done

ksamouchos:
I added a detailed documentation for functions "evalHighOrderProjectionOnLine" and "evalHighOrderProjectionOnTriangle".

ksamouchos:
I'm answering to the question of @andrea-iob about why I removed the assert from function
https://github.com/optimad/bitpit/blob/4040cdd2779366d7605d15582ef7e8f5f031e17c/src/levelset/levelSetSegmentationObject.cpp#L2754

Let's have a look at one of the new integration tests I've added which is not working without removing the assert. The test is the following:
https://github.com/optimad/bitpit/blob/4040cdd2779366d7605d15582ef7e8f5f031e17c/test/integration_tests/levelset/test_levelset_00001.cpp#L183

In this case the size of the narrow band is zero. Thus, function "fillCellGeometricNarrowBandLocationCache" finds valoid cell support only for intersected cells. For the rest returns an  UNKNOWN location:
https://github.com/optimad/bitpit/blob/4040cdd2779366d7605d15582ef7e8f5f031e17c/src/levelset/levelSetSegmentationObject.cpp#L2760

However, when a smoothed geometry is used, there are cells like the one with id=2824 which are intersected by the discretized geometry but not by the smoothed geometry. These cells have a valid cell support, so the function continues its process by checking if the location is NARROW_BAND_INTERSECTED or NARROW_BAND_DISTANCE in line
https://github.com/optimad/bitpit/blob/4040cdd2779366d7605d15582ef7e8f5f031e17c/src/levelset/levelSetSegmentationObject.cpp#L2774

Cells like the one mentioned above don't belong to any of these two categories meaning that the assert that follows will throw an error. By removing the assert, the code still functions normally, because the location of these cells is declared later in function "fillCartesianCellZoneCache". Either these cells are neighboring the narrow band and detected in line
https://github.com/optimad/bitpit/blob/4040cdd2779366d7605d15582ef7e8f5f031e17c/src/levelset/levelSetSegmentationObject.cpp#L2713
or belong to the bulk and are detected in line
https://github.com/optimad/bitpit/blob/4040cdd2779366d7605d15582ef7e8f5f031e17c/src/levelset/levelSetSegmentationObject.cpp#L2726
More specifically, cell 2824 belongs to the second category.

Consequently, the removal of the assert is necessary and does not harm the code.

andrea-iob:
> Consequently, the removal of the assert is necessary and does not harm the code.

Ok. 

Let's avoid caching unknown cell locations (like we already do in ```LevelSetObject::fillCellGeometricNarrowBandLocationCache```) and add a comment explaining that, in the high order case, it's fine to have cells whose cell location will be unknown:
```
    // Update the cell location cache
    //
    // First we need to check if the cell intersects the surface, and only if it doesn't we
    // should check if its distance is lower than the narrow band size.
    //
    // When high order smoothing is active, there may be cells whose support is within the search
    // radius, but they are not intersected and their distance is less than the narrow band size.
    // These cells are not geometrically inside the narrow band, they are neighbours of cells
    // geometrically inside the narrow band and as such it's up to the caller of this function to
    // identify their cell location.
    LevelSetCellLocation cellLocation = LevelSetCellLocation::UNKNOWN;
    if (_intersectSurface(id, cellUnsigendValue, CELL_LOCATION_INTERSECTION_MODE) == LevelSetIntersectionStatus::TRUE) {
        cellLocation = LevelSetCellLocation::NARROW_BAND_INTERSECTED;
    } else if (cellUnsigendValue <= m_narrowBandSize) {
        cellLocation = LevelSetCellLocation::NARROW_BAND_DISTANCE;
    }
    assert((getSurfaceSmoothing() == LevelSetSurfaceSmoothing::HIGH_ORDER) || (cellLocation != LevelSetCellLocation::UNKNOWN));

    if (cellLocation != LevelSetCellLocation::UNKNOWN) {
        CellCacheCollection::ValueCache<char> *locationCache = getCellCache<char>(m_cellLocationCacheId);
        locationCache->insertEntry(id, static_cast<char>(cellLocation));
    }
```

andrea-iob (approved):

andrea-iob:
If you address the last two comments the pull request can be merged.

src/patchkernel/surface_kernel.hpp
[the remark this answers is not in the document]
ksamouchos: Done

ksamouchos:
> > Consequently, the removal of the assert is necessary and does not harm the code.
> 
> Ok.
> 
> Let's avoid caching unknown cell locations (like we already do in `LevelSetObject::fillCellGeometricNarrowBandLocationCache`) and add a comment explaining that, in the high order case, it's fine to have cells whose cell location will be unknown:
> 
> ```
>     // Update the cell location cache
>     //
>     // First we need to check if the cell intersects the surface, and only if it doesn't we
>     // should check if its distance is lower than the narrow band size.
>     //
>     // When high order smoothing is active, there may be cells whose support is within the search
>     // radius, but they are not intersected and their distance is less than the narrow band size.
>     // These cells are not geometrically inside the narrow band, they are neighbours of cells
>     // geometrically inside the narrow band and as such it's up to the caller of this function to
>     // identify their cell location.
>     LevelSetCellLocation cellLocation = LevelSetCellLocation::UNKNOWN;
>     if (_intersectSurface(id, cellUnsigendValue, CELL_LOCATION_INTERSECTION_MODE) == LevelSetIntersectionStatus::TRUE) {
>         cellLocation = LevelSetCellLocation::NARROW_BAND_INTERSECTED;
>     } else if (cellUnsigendValue <= m_narrowBandSize) {
>         cellLocation = LevelSetCellLocation::NARROW_BAND_DISTANCE;
>     }
>     assert((getSurfaceSmoothing() == LevelSetSurfaceSmoothing::HIGH_ORDER) || (cellLocation != LevelSetCellLocation::UNKNOWN));
> 
>     if (cellLocation != LevelSetCellLocation::UNKNOWN) {
>         CellCacheCollection::ValueCache<char> *locationCache = getCellCache<char>(m_cellLocationCacheId);
>         locationCache->insertEntry(id, static_cast<char>(cellLocation));
>     }
> ```

Done

ksamouchos:
Should I squash the commits before merging?

andrea-iob:
> Should I squash the commits before merging?

Keep the commits separate.

PR #462, opened 2024-04-12
levelset: interface intersection

ksamouchos:
1. Introduce the pixel, triangle and polygon intersection with plane.
2. Allow the intersection of interfaces with smoothed continuous surface. It's mandatory for the conversion of Immerflow to a cut-cell software.

### DIFF
--- a/src/CG/CG_elem.cpp
+++ b/src/CG/CG_elem.cpp
@@ -29,6 +29,7 @@
 
 # include <assert.h>
 
+# include "../patchkernel/element_reference.hpp"
 # include "bitpit_operators.hpp"
 
 # include "CG.hpp"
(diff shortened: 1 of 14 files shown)


ksamouchos:
Rebased on current branch #461

ksamouchos:
Rebased on current master

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -1768,6 +1768,119 @@ int LevelSetSegmentationBaseObject::evalCellPart(long id) const
...
+ * @param[in] tolerance is the tolerance used for distance comparisons
+ * @param[in] signedLevelSet controls if signed levelset function will be used
+ * @param[in] tolerance is the tolerance used for distance comparisons
+ * @param[out] centroid is the centroid the intersected face computed as the arithmetic avretage of the face vertiecs
+ * @param[out] area is the area of the intersected face
+ * @param[out] intersectionCentroid is the centroid of the intersection computed as the arithmetic avretage of the intersection points
+ * intersection points. If the interface is not intersected, the maximum numeric limit is returned
+ * @return indicator regarding intersection
+ */
+LevelSetIntersectionStatus LevelSetSegmentationBaseObject::_evalIntersectedInterfaceData(long id, int intrLocalId, bool signedLevelSet, double tolerance, std::array<double, 3> *centroid, double *area, std::array<double, 3> *intersectionCentroid) const
```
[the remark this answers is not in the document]
andrea-iob: Could this function return only the intersection points and leave the task of evaluating other information to the caller (it can be an overload of _intersectInterfaceSurface)? I mean, centroid and intersectionCentroid seems very specific to the application you are developing (i.e., cut cell method in fluid-dynamic solver). The levelset module doesn't seem the correct place for evaluating this information. The levelset should only provide the intersection between a cell interface and the levelset-zero.

src/levelset/levelSetObject.hpp
```diff
@@ -90,7 +90,7 @@ friend class LevelSetProxyObject;
     virtual bool                                isCellInNarrowBand(long id) const;
     virtual bool                                isInNarrowBand(const std::array<double,3> &point) const;
 
-    LevelSetIntersectionStatus                  intersectSurface(long, LevelSetIntersectionMode=LevelSetIntersectionMode::FAST_FUZZY) const;
+    LevelSetIntersectionStatus                  intersectCellSurface(long, LevelSetIntersectionMode=LevelSetIntersectionMode::FAST_FUZZY) const;
```
andrea-iob, line 93: This is a public function, it may be used by some external application using bitpit. You need to add the function with the new name and deprecated the old one.

andrea-iob: The original name was bad, the new one it's not better. Since we are renaming the function, can we find a better name, for example isCellIntersected?

src/levelset/levelSetObject.hpp
```diff
@@ -179,6 +179,7 @@ friend class LevelSetProxyObject;
     virtual void                                dump(std::ostream &);
     virtual void                                restore(std::istream &);
 
+    virtual LevelSetIntersectionStatus          _intersectInterfaceSurface(long id, int intrLocalId, double tolerance) const;
```
andrea-iob, line 182: All these new functions receive in input and interface. However, we are passing the id of a cell and a local id of the interface. Can we just pass the id of the interface?

andrea-iob (changes requested):

src/CG/CG_elem.cpp
```diff
@@ -2577,6 +2577,217 @@ bool intersectPointBox( array3D const &P, array3D const &B1, array3D const &B2,
...
+    return -1;
+}
+
+/*!
+ * Checks if plane intersects a box face and computes the nodes comprising the intersected
+ * face belonging to the subspace opposite of the one pointed by the plane's normal
+ *
+ * In the degenerate case where the box face belongs to the given plane the interface
+ * is not considered as intersected.
+ *
```
andrea-iob, line 2619: This seems inconsistent with all other functions in this module.

src/levelset/levelSetKernel.cpp
```diff
@@ -171,6 +171,79 @@ bool LevelSetKernel::intersectCellPlane( long id, const std::array<double,3> &ro
     return CGElem::intersectPlaneBox( root, normal, minPoint, maxPoint, dim, tolerance);
 }
 
+/*!
+ * Given a cell considered as a box, and one of its interfaces, considered as one fo the six box
```
andrea-iob, line 175: From the documentation it's not clear what this function does. How can you consider a tetrahedron a box?

src/levelset/levelSetKernel.cpp
```diff
@@ -171,6 +171,79 @@ bool LevelSetKernel::intersectCellPlane( long id, const std::array<double,3> &ro
...
+ * @param[out] intersectionCentroid is the centroid of the intersection computed as the arithmetic avretage of the intersection points
+ * @return true if interface is intersected
+ */
+bool LevelSetKernel::intersectInterfacePlane( long id, int intrLocalId, const std::array<double,3> &root, const std::array<double,3> &normal, double tolerance, std::array<double, 3> *interfaceCentroid, double *interfaceArea, std::array<double, 3> *intersectionCentroid ) {
+
+    int boxFaceId = computeBoxFaceIndex(id, intrLocalId, tolerance);
+
+    std::array<double,3> minPoint;
+    std::array<double,3> maxPoint;
+    m_mesh->evalCellBoundingBox(id, &minPoint, &maxPoint);
```
andrea-iob, line 241: It seems it would be easier to just intersect the interface with the plane, without going through the bounding box. It just an intersection between a polygon and a plane, there may be better algorithm but a simple one should be: loop though the sides of the polygon, discard the sides that have both vertices below the plane, keep the sides that have both vertices above the plane, evaluate the intersection point if the side has one vertex above and one below the plane. Am I missing something?

andrea-iob: In CG there can be a set of function that does the cut: cutLine, cutTriangle, cutPolygon. These functions take as input the shape (line, triangle, polygon, ...) and the plane and return a list of points that define the shape after the cut.

src/CG/CG_elem.cpp
```diff
@@ -2577,6 +2577,217 @@ bool intersectPointBox( array3D const &P, array3D const &B1, array3D const &B2,
...
+ * \param[in] n normal to plane
+ * \param[in] A0 min point of box
+ * \param[in] A1 max point of box
+ * \param[in] dim number of dimensions
+ * \param[in] distanceTolerance if distance among features exceed this value they are considered as not intersecting
+ * \param[out] interfacePoints pointed vector holds the vertices defining the face intersection in counterclockwise order
+ * \param[out] intersectionPoints pointed vector holds the vertices of the face resulted from its intersection in counterclockwise order
+ * \return if intersects
+ */
+bool intersectPlaneBoxFace(int i, array3D const &P, array3D const &n, array3D const &A0, array3D const &A1, int dim, const double distanceTolerance, std::vector<array3D> *interfacePoints, std::vector<array3D> *intersectionPoints)
```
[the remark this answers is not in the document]
ksamouchos: Done

src/CG/CG_elem.cpp
```diff
@@ -2577,6 +2577,217 @@ bool intersectPointBox( array3D const &P, array3D const &B1, array3D const &B2,
...
+ * \param[in] A0 min point of box
+ * \param[in] A1 max point of box
+ * \param[in] dim number of dimensions
+ * \param[in] distanceTolerance if distance among features exceed this value they are considered as not intersecting
+ * \param[out] faceCentroid is the centroid of the intersected face computed as the arithmetic avretage of the face vertiecs
+ * \param[out] faceArea is the area of the intersected face
+ * \param[out] intersectionCentroid is the centroid of the intersection computed as the arithmetic avretage of the intersection points
+ * \return true if interface is intersected
+ */
+bool computeIntersectedBoxFaceData(int i, array3D const &P, array3D const &n, array3D const &A0, array3D const &A1, int dim, const double distanceTolerance, std::array<double, 3> *faceCentroid, double *faceArea, std::array<double, 3> *intersectionCentroid)
```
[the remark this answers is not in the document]
ksamouchos: I removed it

src/levelset/levelSetKernel.cpp
```diff
@@ -171,6 +171,79 @@ bool LevelSetKernel::intersectCellPlane( long id, const std::array<double,3> &ro
...
+ * Given a cell considered as a box, and one of its interfaces, considered as one fo the six box
+ * interfaces, it computes the index corresponding to that box face based on the enumeration
+ * defined inCGElem namespace.
+ * The normal to the plane points towards the subspace.
+ *
+ * @param[in] id is the index of cell
+ * @param[in] intrLocalId is the local index of the interface for the given cell
+ * @return the index corresponding to the given box face 
+ */
+int LevelSetKernel::computeBoxFaceIndex( long id, int intrLocalId, double tolerance) {
```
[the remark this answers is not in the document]
ksamouchos: I removed it

src/levelset/levelSetKernel.cpp
```diff
@@ -171,6 +171,79 @@ bool LevelSetKernel::intersectCellPlane( long id, const std::array<double,3> &ro
...
+ * @param[in] intrLocalId is the local index of the interface for the given cell
+ * @param[in] root is a point on the plane
+ * @param[in] normal is the normal of the plane
+ * @param[in] tolerance is the tolerance used for distance comparisons
+ * @param[out] interfaceCentroid is the centroid the intersected face computed as the arithmetic avretage of the face vertiecs
+ * @param[out] interfaceArea is the area of the intersected face
+ * @param[out] intersectionCentroid is the centroid of the intersection computed as the arithmetic avretage of the intersection points
+ * @return true if interface is intersected
+ */
+bool LevelSetKernel::intersectInterfacePlane( long id, int intrLocalId, const std::array<double,3> &root, const std::array<double,3> &normal, double tolerance, std::array<double, 3> *interfaceCentroid, double *interfaceArea, std::array<double, 3> *intersectionCentroid ) {
```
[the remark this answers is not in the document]
ksamouchos: It is relocated to the PatchKernel module

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -1768,6 +1768,119 @@ int LevelSetSegmentationBaseObject::evalCellPart(long id) const
...
+{
+    // Evaluate projection of interface centroid on surface
+    long intrGlobalId = m_kernel->getMesh()->getCell(id).getInterfaces()[intrLocalId];
```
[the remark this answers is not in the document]
ksamouchos: The transition from the "global" to the "local" id is not needed anymore. The id of the interface is passed as an argument.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -1768,6 +1768,119 @@ int LevelSetSegmentationBaseObject::evalCellPart(long id) const
...
+            (*centroid)             = interfaceCentroid;
+            (*area)                 = m_kernel->getMesh()->evalInterfaceArea(intrGlobalId);
+            (*intersectionCentroid) = {maxNum, maxNum, maxNum};
+            return LevelSetIntersectionStatus::FALSE;
+        }
+    }
+    double maxNum           = std::numeric_limits<double>::max();
+    (*centroid)             = interfaceCentroid;
+    (*area)                 =  0.0;
+    (*intersectionCentroid) = {maxNum, maxNum, maxNum};
```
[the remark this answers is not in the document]
ksamouchos: Done

src/levelset/levelSetObject.hpp
ksamouchos: Done

src/levelset/levelSetObject.hpp
ksamouchos: Done

src/CG/CG_elem.cpp
ksamouchos: I kept this change. All the functions I added for intersecting pixels, triangles, and polygons are compatible with each other.

src/levelset/levelSetKernel.cpp
ksamouchos: This function is removed.

src/levelset/levelSetKernel.cpp
ksamouchos: Done. I used already implemented functions whenever possible.

ksamouchos:
After the review of @andrea-iob and the discussion I had with @marcocisternino and @andrea-iob I completely change this branch by reprogramming everything from scratch. New features are introduced and others are removed. The new version is more generic and does not depend exclusively on my work related to the cut-cell method. This branch is responsible for:

1. the introduction of the pixel, triangle and polygon intersection with planes
2. the intersection of interfaces with smoothed continuous surfaces

The considerable amount of modifications requires a new review to the whole pull-request.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -3151,8 +3152,14 @@ void LevelSetSegmentationObject::_evalProjection(const std::array<double,3> &poi
...
     // and in order to have a normal that is agnostic with respect the two sides of the surface.
+    // The sign computation requires the projection reevaluation!
     if (!signedLevelSet) {
-        (*projectionNormal) *= static_cast<double>(evalSign(point));
+        short sign = evalSign(point);
+        if (sign == 0) {
+            throw std::runtime_error("Cannot evaluate the signed projection becuase the given point lays on the zero-level-set surface");
+        } else {
+            (*projectionNormal) *= static_cast<double>(sign);
+        }
```
andrea-iob, line 3248: Don't throw an exception, otherwise the function will not be able to evaluate the projection of points on the zero-level-set surface, which can be performed without problems if the normal is not needed. Just do two changes:
 - allow to pass null pointers as arguments and perform the needed calculation only when the needed pointer are not null;
 - update the documentation of the function:
 ```
* \param[out] projectionPoint if a valid pointer is provided, out output it will contain the coordinates of the projection point on the surface
 * \param[out] projectionNormal if a valid pointer is provided, out output it will contain the normal to the surface on the projection point. If the projection point lies on the zero-level-set surface, and empty vector is returned
```

Comment (andrea-iob) [reply src/levelset/levelSetSegmentationObject.cpp:3248]: Also, maybe find a better commit message.

File: src/patchkernel/patch_kernel.cpp [hunk 14]
```diff
@@ -5635,6 +5635,147 @@ void PatchKernel::getInterfaceVertexCoordinates(long id, std::array<double, 3> *
...
+   opposite to the one pointed by the given normal. If the interface is not intersected,
+   an empty polygon is returned. 
+
+   Interfaces laid on the plane are not considered as intersected ones.
+
+   Interfaces having an edge on the plane are not considered as intersected ones.
+
+   \param[in] P0 First triangle vertex
+   \param[in] P1 Second triangle vertex
+   \param[in] P2 Third triangle vertex
```
andrea-iob, line 5650: The documentation doesn't match the argument of the function.

src/patchkernel/patch_kernel.cpp
```diff
@@ -5635,6 +5635,147 @@ void PatchKernel::getInterfaceVertexCoordinates(long id, std::array<double, 3> *
...
+   \param[in] P Point belonging to the plane
+   \param[in] nP Normal to the plane
+   \param[out] intersection If intersects, return the cutting segment. In case of a liner interface
+   the intersection degenerates to a point which coincides with the begining and the ending
+   of the segment
+   \param[out] polygon is the polygon constructed by the interface intersection placed on the
+   opposite subspace of the one pointed by the plane's normal
+   \param[in] distanceTolerance define a distance tolerance so that if a polygon and the 
+   plane are at distance below such threshold they are considered intersecting.
```
andrea-iob, line 5659: Not needed.

src/patchkernel/patch_kernel.cpp
```diff
@@ -5635,6 +5635,147 @@ void PatchKernel::getInterfaceVertexCoordinates(long id, std::array<double, 3> *
...
+		std::array<double, 2> A1 = {maxPoint[d0], maxPoint[d1]};
+
+		return CGElem::intersectPlanePixel(P, nP, A0, A1, dir, constCoor,
+                       *intersection, *polygon, distanceTolerance);
+	}
+
+	default:
+	{
+	        ConstProxyVector<long> vertexIds = interface.getVertexIds();
+                int nVertices = vertexIds.size();
```
andrea-iob, line 5755: You should get the vertices in counter clockwise order:
```
assert(interfaceType != ElementType::UNDEFINED);
const Reference2DElementInfo &interfaceInfo = static_cast<const Reference2DElementInfo &>(interfaceType);

std::vector<std::array<double, 3>> coordinates(nVertices)
for (std::size_t i = 0; i < nVertices; ++i) {
    long vertexId = vertexIds[interfaceInfo.getCCWOrderedVertex(i)]
    coordinates[i] = getVertex(vertexId).getCoords();
}
```

or add to PatchKernel a function like "ConstProxyVector<long> VolumeKernel::getFaceOrderedVertexIds(const Cell &cell, int face) const" that takes in input an interface.

src/patchkernel/patch_kernel.cpp
```diff
@@ -5635,6 +5635,147 @@ void PatchKernel::getInterfaceVertexCoordinates(long id, std::array<double, 3> *
...
+
+		std::unique_ptr<std::array<double, 3>[]> coordinates;
```
andrea-iob, line 5757: Don't use std::unique_ptr, a vector is fine here.

src/patchkernel/patch_kernel.cpp
```diff
@@ -5635,6 +5635,147 @@ void PatchKernel::getInterfaceVertexCoordinates(long id, std::array<double, 3> *
 	getElementVertexCoordinates(interface, coordinates);
 }
 
+/*!
+   Check if an interface  is intersected by a given plane. Calculate the intersection
+   as a segment defined by two points and the part of the interface belonging to the subspace
+   opposite to the one pointed by the given normal. If the interface is not intersected,
+   an empty polygon is returned. 
+
+   Interfaces laid on the plane are not considered as intersected ones.
```
andrea-iob, line 5644: I still don't like this, because it seems counter intuitive.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -1817,6 +1817,92 @@ LevelSetIntersectionStatus LevelSetSegmentationBaseObject::_isCellIntersected(lo
...
+ * The iso-surface is considered planar. Calculate the intersection
+ * as a segment defined by two points and the part of the interface belonging to the subspace
+ * of negative level-set. If the interface is not intersected, an empty polygon is returned. 
+ * The interface belonging to the possitive level-set subspace is returned if the
+ * "positivePart" argument is true.
+ *
+ * The intersection is computed by considering the zero-levelset as a curved surface.
+ *
+ * @param[in] id is the interface index
+ * @param[in] positivePart when true the part of the  interface occupying the positive levelset area is returned
```
andrea-iob, line 1832: This seems an argument to decide which portion of the intersect interface should be returned. It that's the case, we could mimic what ParaView does: let's define a default behaviour and then pass an argument called "invert" which inverts which part of the intersected interfae is considered.

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -1817,6 +1817,92 @@ LevelSetIntersectionStatus LevelSetSegmentationBaseObject::_isCellIntersected(lo
...
+ * @param[in] tolerance is the tolerance used for distance comparisons
+ * @param[out] intersection If intersects, return the cutting segment. In case of a liner interface
+ * the intersection degenerates to a point which coincides with the begining and the ending
+ * of the segment
+ * @param[out] polygon is the polygon constructed by the interface intersection placed on the
+ * opposite subspace of the one pointed by the plane's normal
+ * @return the level set status
+ */
+LevelSetIntersectionStatus LevelSetSegmentationBaseObject::_isInterfaceIntersected(long id, bool positivePart,
+                                                                                   double tolerance,
```
andrea-iob, line 1842: Get the tolerance using the levelset function "double LevelSetKernel::getDistanceTolerance() const".

src/patchkernel/patch_kernel.cpp
```diff
@@ -5635,6 +5635,147 @@ void PatchKernel::getInterfaceVertexCoordinates(long id, std::array<double, 3> *
...
+   \return True if intersection occurs, false if not
+ */
+bool PatchKernel::intersectInterfacePlane(long id, std::array<double, 3> P, std::array<double, 3> nP,
+                                          std::array<std::array<double, 3>, 2> *intersection,
+                                          std::vector<std::array<double, 3>> *polygon) const
+{
+        double distanceTolerance = getTol();
+
+	const Interface &interface = getInterface(id);
+	ElementType elementType = interface.getType();
```
andrea-iob, line 5669: interfaceType

src/patchkernel/patch_kernel.cpp
```diff
@@ -5635,6 +5635,147 @@ void PatchKernel::getInterfaceVertexCoordinates(long id, std::array<double, 3> *
...
+                                 intersections, polygons, distanceTolerance);
+
+		if (intersections.size() > 1) {
+			throw std::runtime_error("Intersection of concave interface is not supported.");
+                }
+
+		(*intersection) = intersections[0];
+		(*polygon)      = polygons[0];
+
+		return intrExist;
```
andrea-iob, line 5772: intersectionFound

src/patchkernel/patch_kernel.cpp
```diff
@@ -5635,6 +5635,147 @@ void PatchKernel::getInterfaceVertexCoordinates(long id, std::array<double, 3> *
...
+		std::unique_ptr<std::array<double, 3>[]> coordinates;
+	        getVertexCoords(nVertices, vertexIds.data(), &coordinates);
+
+                std::vector<std::array<std::array<double, 3>, 2>> intersections;
+                std::vector< std::vector<std::array<double, 3>>> polygons;
+		bool intrExist = CGElem::intersectPlanePolygon(P, nP, nVertices, coordinates.get(),
+                                 intersections, polygons, distanceTolerance);
+
+		if (intersections.size() > 1) {
+			throw std::runtime_error("Intersection of concave interface is not supported.");
```
andrea-iob, line 5766: interface -> interfaces

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+ * Given the normal cartesian direction dir, the pixel coordinates
+ * are represened on its 2D plane as (x, y) where x = (dim + 1) % 3 and y = (dim + 2) % 3 
+ *
+ * Local indeces for edges are defined as follows:
+ * edge 0: y = y min
+ * edge 1: x = x max
+ * edge 2: y = y max
+ * edge 3: x = x min
+ *
+ * Co-planarity of input pixel and plane will produce no intersection.
```
andrea-iob, line 2131: This seems not coherent with all other functions in this module.

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+ * No intersection is considered if the plane passes from a pixel's vertex,
+ * and no other intersection exist.
+ *
+ * \param[in] Pp point belonging to the plane
+ * \param[in] nP unit normal to plane
+ * \param[in] A0 min point of pixel on the 2D space defined by the pixel direction dir
+ * \param[in] A1 max point of pixel on the 2D space defined by the pixel direction dir
+ * \param[in] dir is an integer from 1 to 3 and indicates the cartesian direction normal
+ * to the pixel. Its sign affects the polynomial's orientation. The above used dim variable
+ * is computed as dim = |dir| - 1
```
andrea-iob, line 2141: What's the point of this argument? You have the coordinates of the vertices of the pixel ordered in a well defined way, this should be enough.

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+ * \param[in] coor indicates the cartesian coordinate of all the pixel points
```
andrea-iob, line 2142: Document the order in which the vertex coordinates should be passed.

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
     return false;
 }
 
+/*!
+ * Check if a pixel is crossing a given plane and fully calculate the intersection (a segment) with 
+ * edges, eventually. A pixel is defined as a planar rectangular, vertical to a cartesian axis.
```
andrea-iob, line 2118: As far as I know, there is no requirement for the normal of the plane in which the pixel lies to be aligned with one of the Cartesian axes.

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+ * \param[in] coor indicates the cartesian coordinate of all the pixel points
+ * \param[out] intersection If intersects, return the cutting segment
+ * \param[out] edges_of_intersection If intersects, return the pixel edge local indices, 
+ * of the intersected edges. If no intersection occurs return a -1 label. 
+ * \param[in] distanceTolerance if distance among features exceed this value they are considered as not intersecting
+ * \return if intersects
+ */
+bool intersectPlanePixel(array3D const &Pp, array3D const &nP, std::array<double, 2> const &A0,
+                         std::array<double, 2> const &A1, int dir, double coor, std::array<array3D, 2> &intersection,
+                         std::array<int, 2> &edges_of_intersection, const double distanceTolerance)
```
andrea-iob, line 2151: Is this function much faster than intersectPlanePolygon? In the CG module this will be the only function that deals with pixels. We can just drop it and call intersectPlanePolygon for pixels too.

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+void reconstructPlaneIntersectedPolygons(std::size_t nV, array3D const *V,
+                                         std::vector<std::array<array3D, 2>> const &Qs,
+                                         std::vector<std::array<int, 2>> const &edges_of_Qs,
+                                         std::vector< std::vector<array3D> > &polys,
+                                         const double distanceTolerance)
+{
+    polys.clear();
+
+    std::vector<array3D> motherPolygon;
+    std::size_t motherPos(0);
```
andrea-iob, line 2870: std::size_t motherPos = 0;
int motherInit = -1;

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+            continue;
+        }
+
+        const int &L1     = edges[0];
+        const int &L2     = edges[1];
+        const array3D &S1 = Q[0];
+        const array3D &S2 = Q[1];
+
+        //if segment has negligible lenght
+        if (norm2(S1 - S2) < distanceTolerance) {
```
andrea-iob, line 2890: You can use ```if (utils::DoubleFloatingEqual()(S1, S2, distanceTolerance )) {```.

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+                                         const double distanceTolerance)
+{
+    polys.clear();
+
+    std::vector<array3D> motherPolygon;
+    std::size_t motherPos(0);
+    int motherInit(-1);
+    motherPolygon.reserve(nV);
+
+    std::size_t size = Qs.size();
```
andrea-iob, line 2874: Please add some comments explaining what motherPolygon is.

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+            subTrianglesCount = 1;
+            subTrianglesConn  = {{0, 1, 2}};
+        } else {
+            subTrianglesCount = nV;
+            subTrianglesConn.resize(3 * nV);
+
+            for (std::size_t i = 0; i < nV; ++i) {
+                subTrianglesConn[3 * i]     = i;
+                subTrianglesConn[3 * i + 1] = (i + 1) % nV;
+                subTrianglesConn[3 * i + 2] = -1;
```
andrea-iob, line 2710: Are you triangulating the polygon? The module CG has functions to do that: subtriangleOfPolygon, polygonSubtriangleCount.

src/CG/CG_elem.cpp
```diff
@@ -2113,6 +2113,829 @@ bool intersectSegmentPolygon( array3D const &P0, array3D const &P1, std::size_t
...
+ * \param[in] C Centroid of the polygon used as an auxxiliary point during the triangulation
+ * \param[in] subTrianglesCount number of triangles constituting the polygon
+ * \param[in] subTrianglesConn connectivity of each triangle being part of the triangulated polygon
+ * \param[out] Qs If intersects, return a non-empty list of intersecting Segments.
+ * \param[out] edges_of_Qs If intersects, return the list of polygon edges couple "touched" by the intersecting segment, one for each
+ * intersecting segment.
+ * \param[in] distanceTolerance if distance among features exceed this value they are considered as not intersecting
+ * \return True if intersect, false if not. Return false too if class is not updated.
+ */
+bool intersectPlaneTriPolygon(array3D const &P, array3D const &nP, bool intersect,
```
andrea-iob, line 2739: Do you really need this function? It doesn't seem very efficient to triangulate a polygon and then call this function. You can just evaluate the triangulation in place when you need it:
```
    int triangleCount = polygonSubtriangleCount(nV, V);
    for (int triangle=0; triangle < triangleCount; ++triangle) {
        subtriangleOfPolygon( triangle, nV, V, V0, V1, V2);

        ...
    }
```
see for example projectPointPolygon.

andrea-iob (changes requested):

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -3121,6 +3121,7 @@ long LevelSetSegmentationObject::_evalSupport(const std::array<double,3> &point,
  * \param[in] point are the coordinates of the given point
  * \param support is the the closest segment to the specified point
  * \param[in] signedLevelSet controls if signed levelset function will be used
+ * A false value requires the projection evaluation twice
```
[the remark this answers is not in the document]
ksamouchos: Documentation corrected

src/levelset/levelSetSegmentationObject.cpp
```diff
@@ -3151,8 +3152,14 @@ void LevelSetSegmentationObject::_evalProjection(const std::array<double,3> &poi
 
     // If an unsigned evaluation is requested, the orientation of the surface should be discarded
     // and in order to have a normal that is agnostic with respect the two sides of the surface.
+    // The sign computation requires the projection reevaluation!
```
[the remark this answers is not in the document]
ksamouchos: Done

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done

src/patchkernel/patch_kernel.cpp
ksamouchos: Done

src/patchkernel/patch_kernel.cpp
ksamouchos: Done by using the code you provided

src/patchkernel/patch_kernel.cpp
ksamouchos: Done

src/patchkernel/patch_kernel.cpp
ksamouchos: If the interface lays on the plane, the intersection becomes the face itself. However, this function is designed to provide a segment as an intersection. Also, if we consider this case as an intersection, we have to consider others as well, like when the interface has only one of its vertices on the plane. In this case the intersection is just this node. Finally, in a case when an edge is on the surface, the user won't be able to understand without further computations if the intersection is a "real" one or just the edge of the interface, because in both cases the intersection will be represented by a segment. Are you sure that you prefer this behavior?

src/levelset/levelSetSegmentationObject.cpp
ksamouchos: Done

src/patchkernel/patch_kernel.cpp
ksamouchos: Done

src/patchkernel/patch_kernel.cpp
ksamouchos: Done

src/patchkernel/patch_kernel.cpp
ksamouchos: Done

src/CG/CG_elem.cpp
ksamouchos: See my comment in a previous similar case.

src/CG/CG_elem.cpp
ksamouchos: It was needed, because the vertices weren't given as arguments to the functions. This has changed. By giving the vertices coordinates, most of those arguments are not needed any more.

src/CG/CG_elem.cpp
ksamouchos: This argument is not needed anymore.

src/CG/CG_elem.cpp
ksamouchos: Done. Now this function handles pixels of an arbitrary orientation.

src/CG/CG_elem.cpp
ksamouchos: It's 1.3 times faster. For me, this number is enough to keep the handling of pixels separately. This function is going to be called thousands of times during the cut-cell mesh generation. Usually, this is one of the slowest parts during this process and must be as fast as possible. If you prefer, I can add intersections of pixels with other geometrical elements, so the library will be more complete.

src/CG/CG_elem.cpp
ksamouchos: Done

src/CG/CG_elem.cpp
ksamouchos: Done

src/CG/CG_elem.cpp
ksamouchos: This is a function written by Rocco. He didn't use the triangulation of bitpit on purpose, because it couldn't handle all the possible cases like concave polygons if I'm not mistaken.

src/CG/CG_elem.cpp
ksamouchos: Rocco didn't use the triangulation of bitpit on purpose, because it couldn't handle all the possible cases like concave polygons if I'm not mistaken.

ksamouchos:
I updated the branch following the review of @andrea-iob . I believe that @edoardolombardi and @roccoarpa should also be added as reviewers, because this pull-request affects their work.

roccoarpa:
hello @ksamouchos, thank you for the warning, but for me it's difficult to do a proper review since there's lot of stuff going on in this pull (not only the part that interest me). This is my personal opinion: for me it's ok whatever you decide to do, as long as the code works and it's properly tested. I will update my part in mimic eventually. Ty.

ksamouchos:
> hello @ksamouchos, thank you for the warning, but for me it's difficult to do a proper review since there's lot of stuff going on in this pull (not only the part that interest me). This is my personal opinion: for me it's ok whatever you decide to do, as long as the code works and it's properly tested. I will update my part in mimic eventually. Ty.

ok, thank you for your kind response

src/CG/CG_elem.cpp
andrea-iob: Both triangulations do the same thing: each sub-triangle is defined by two consecutive vertices and the centroid:
```
                subTrianglesConn[3 * i]     = i;
                subTrianglesConn[3 * i + 1] = (i + 1) % nV;
                subTrianglesConn[3 * i + 2] = -1;
```
with ```-1``` being replaced by the centroid of the polygon.

```
    V0 = {0., 0., 0.};
    for (std::size_t i = 0; i < nV; i++) {
        for (int d = 0; d < 3; d++) {
            V0[d] += V[i][d];
        }
    }
    for (int d = 0; d < 3; d++) {
        V0[d] /= nV;
    }
    V1 = V[triangle%nV];
    V2 = V[(triangle+1)%nV];
```
with ```V0``` being explicitly calculated as the centroid of the polygon.

This kind of sub-triangulation can fail for concave polygons.

ksamouchos: I simplified these methods. Now the triangulation is done under the hood and there's no need for calling polygonSubtriangleCount.

andrea-iob (approved):

ksamouchos:
> The changes are fine, however explicitly using the triangulation functions of the CG module would have been better because if these function are improved (for example adding support for concave polygons), also the intersection functions will befit of this improvements for free.

The algorithm works fine for convex or concave polygons. The triangulation is just a trick for using the line-triangle intersection provided by bitpit and it doesn't have any important role in the geometrical construction. That's why it's removed without harming the generality of the algorithm.

ksamouchos:
@edoardolombardi would you like to have a look before merging this branch?

ksamouchos:
One last thing. I don't really like the introduction of function "computePolygonNormal" in CG.hpp. The computation of the polygon's normal already exist in module "patchkernel", but it has a different definition. I think that using the already existing function instead of "computePolygonNormal" won't harm the algorithm, but I'm not completely sure. That's why I kept Rocco's definition.

Let me know, if you have any better idea about how to improve this part.

ksamouchos:
@andrea-iob , @marcocisternino and @edoardolombardi if you don't have any other comments to make, I'll merge this branch tomorrow morning

PR #464, opened 2024-04-22
discretization: optimize evaluation of gradient weights

andrea-iob:
Instead of performing multiple matrix-vector multiplications it is possible to perform a single matrix-matrix multiplication.


marcocisternino (approved):
code checked and tested using gradients in several cases of an application. No changes in application results.

PR #470, opened 2024-06-06
patchkernel: preliminary changes for tracking adaptation of interface/vertex

andrea-iob:
Mainly some changes to the data structure that keeps track of element changes that will make it easier to add support for tracking adaptation of interface/vertex.


edoardolombardi (approved):

marcocisternino (approved):
Code checked and test outside

PR #472, opened 2024-06-07
common: add a custom implementation of std::void_t for C++11 backwards compatibility

andrea-iob:
Since void_t is the only C++17 feature used in bitpit, this change will allow to keep C++11 compatibility.


andrea-iob:
Backwards compatibility with C++11 and older version of bitpit will be kept only for the incoming v1.11 release. These changes will not be merged into master.

PR #473, opened 2024-06-07
LA: fix initialization of PETSc KSP

andrea-iob:
KSP/PC options should be set before calling KSPSetFromOptions/PCSetFromOptions.


andrea-iob:
This should be backported to bitpit 1.11.

marcocisternino (approved):

PR #475, opened 2024-06-21
 volcartesian: properly update patch data structures when switching memory mode

andrea-iob:
After switching to normal memory mode, we need to update the patch to make sure all data structures (e.g. interfaces) are properly updated.


ksamouchos (approved):
I agree with the changes. I tested it and I don't see the error anymore.

PR #478, opened 2025-03-18
cmake: rework generation of bitpit CMake configuration files

andrea-iob:
Let CMake generate its configuration files. To also allows to simplify how external dependencies are added to the bitpit targets.


marcocisternino (approved):
Tested with mahindra project

